PR #361, opened 2023-07-19
Add accio metadata schema config

brandboat:


accio-main/src/main/java/io/accio/main/pgcatalog/table/PgProcTable.java
```diff
@@ -83,7 +84,7 @@ protected Map<String, String> createTableContent()
         return ImmutableMap.<String, String>builder()
                 .put("oid", "${hash}(concat('PROC', ${functionName}))")
                 .put("proname", "${split}(${functionName}, '__')${firstOrdinal}")
-                .put("pronamespace", "${hash}(${functionSchema})")
+                .put("pronamespace", "${hash}('" + PG_CATALOG + "')")
```
goldmedal, line 87: It should be the pg_catalong name in cnofig.

brandboat: Yeah... that would be ideally, but bi tool / pg jdbc driver only understand the name `pg_catalog`. In this pr what I have changed is the name of `pg_catalog` stored in bigquery is modifiable, while all content in pg_tables should all accept that they are in system schema `pg_catalog`.

goldmedal (approved):
LGTM

PR #365, opened 2023-07-21
Add and Fix PostgreSQL datasource type test

AShiou:
## Description
The mainly is to add PostgreSQL datasource type test, but also fix the following things to make type test working.

What we do,
1. Remove `tinyint` type test, because there is no `tinyint` type in PostgreSQL
2. Fix PostgreSQL connector type
    - `bytea` type
    - `timestamp` type
    - `varchar` type with `unicode`

## Other issue
We can improve the type test by modify expected result. https://github.com/Canner/accio/issues/363

### DIFF
--- a/accio-main/src/main/java/io/accio/main/connector/bigquery/BigQueryRecordIterator.java
+++ b/accio-main/src/main/java/io/accio/main/connector/bigquery/BigQueryRecordIterator.java
@@ -25,6 +25,7 @@ import io.accio.connector.bigquery.BigQueryType;
 import org.apache.commons.lang3.StringUtils;
 import org.joda.time.Period;
 
+import java.time.Instant;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.util.Iterator;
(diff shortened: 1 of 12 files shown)


accio-main/src/main/java/io/accio/main/connector/bigquery/BigQueryRecordIterator.java
```diff
@@ -116,9 +117,11 @@ private static Object getFieldValueByType(Field field, FieldValue fieldValue)
             case DATE:
                 return LocalDate.parse((String) fieldValue.getValue());
             case DATETIME:
-                return convertToMicroseconds(LocalDateTime.parse(fieldValue.getStringValue()));
+                return LocalDateTime.parse(fieldValue.getStringValue());
             case TIMESTAMP:
-                return fieldValue.getTimestampValue();
+                long microSeconds = fieldValue.getTimestampValue();
+                Instant instant = Instant.ofEpochSecond(microSeconds / 1000000, microSeconds % 1000000 * 1000);
```
goldmedal, line 123: Curiously, why  `microSeconds % 1000000 * 1000`? why do we need to adjust the nano second?

goldmedal (approved):
One question. Others LGTM.

PR #378, opened 2023-10-25
Remove syntactic sugar and add to-1 relation support in model expression

brandboat:
reopen https://github.com/Canner/accio/pull/374 as we revert this pr.


goldmedal (approved):
LGTM

PR #379, opened 2023-10-26
Support model on model

brandboat:
## Description
Support model based on model

baseModel field is added in this pr
```json
{
   "name":"BookReplica",
   "refSql":null,
   "baseModel":"Book",
   "columns":[
      {
         "name":"authorId",
         "type":"STRING",
         "relationship":{
            "empty":true,
            "present":false
         }
      }
   ]
}
```


goldmedal (approved):
LGTM

PR #390, opened 2023-11-15
Generate cache data on a per-table basis

goldmedal:
# What's Changed
- generate cache data on a per-table basis
- create the caching task for each dataset(model & metric)

### DIFF
--- a/accio-cache/src/main/java/io/accio/cache/CacheManager.java
+++ b/accio-cache/src/main/java/io/accio/cache/CacheManager.java
@@ -113,18 +113,18 @@ public class CacheManager
             throw new AccioException(GENERIC_USER_ERROR, format("cache is already running; catalogName: %s, schemaName: %s, tableName: %s", mdl.getCatalog(), mdl.getSchema(), cacheInfo.getName()));
         }
         removeCacheIfExist(catalogName, schemaName, tableName);
-        return handleCache(mdl, cacheInfo);
+        return doCache(mdl, cacheInfo);
     }
 
     private CompletableFuture<Void> handleCache(AccioMDL mdl, CacheInfo cacheInfo)
     {
-        return doCache(mdl, cacheInfo)
+        return refreshCache(mdl, cacheInfo)
                 .thenRun(() -> {
                     if (cacheInfo.getRefreshTime().toMillis() > 0) {
                         cacheScheduledFutures.put(
                                 new CatalogSchemaTableName(mdl.getCatalog(), mdl.getSchema(), cacheInfo.getName()),
                                 refreshExecutor.scheduleWithFixedDelay(
-                                        () -> doCache(mdl, cacheInfo).join(),
+                                        () -> createTask(mdl, cacheInfo).join(),
                                         cacheInfo.getRefreshTime().toMillis(),
                                         cacheInfo.getRefreshTime().toMillis(),
                                         MILLISECONDS));
@@ -290,7 +290,7 @@ public class CacheManager
             CatalogSchemaTableName catalogSchemaTableName = new CatalogSchemaTableName(mdl.getCatalog(), mdl.getSchema(), cacheInfo.getName());
             TaskInfo taskInfo = new TaskInfo(mdl.getCatalog(), mdl.getSchema(), cacheInfo.getName(), RUNNING, Instant.now());
             // To fix flaky test, we pass value to tasks instead of a reference;
-            Task task = new Task(TaskInfo.copyFrom(taskInfo), refreshCache(mdl, cacheInfo));
+            Task task = new Task(TaskInfo.copyFrom(taskInfo), handleCache(mdl, cacheInfo));
             tasks.put(catalogSchemaTableName, task);
             return taskInfo;
         });
--- a/accio-cache/src/main/java/io/accio/cache/CacheManager.java
+++ b/accio-cache/src/main/java/io/accio/cache/CacheManager.java
@@ -14,6 +14,7 @@
 
 package io.accio.cache;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import io.accio.base.AccioException;
 import io.accio.base.AccioMDL;
@@ -23,6 +24,7 @@ import io.accio.base.Parameter;
 import io.accio.base.SessionContext;
 import io.accio.base.client.duckdb.DuckdbClient;
 import io.accio.base.dto.CacheInfo;
+import io.accio.base.metadata.SchemaTableName;
 import io.accio.base.sql.SqlConverter;
 import io.accio.cache.dto.CachedTable;
 import io.accio.sqlrewrite.AccioPlanner;
@@ -80,7 +82,7 @@ public class CacheManager
     private final ScheduledThreadPoolExecutor refreshExecutor = new ScheduledThreadPoolExecutor(5, daemonThreadsNamed("cache-refresh-%s"));
 
     private final ExecutorService executorService = newCachedThreadPool(threadsNamed("cache-manager-%s"));
-    private final ConcurrentHashMap<String, Task> tasks = new ConcurrentHashMap<>();
+    private final ConcurrentHashMap<CatalogSchemaTableName, Task> tasks = new ConcurrentHashMap<>();
 
     @Inject
     public CacheManager(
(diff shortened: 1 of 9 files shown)


accio-cache/src/main/java/io/accio/cache/CacheManager.java
```diff
@@ -250,29 +265,38 @@ public void removeTempFile(PathInfo pathInfo)
...
     {
         return supplyAsync(() -> {
-            String taskId = randomUUID().toString();
-            TaskInfo taskInfo = new TaskInfo(taskId, mdl.getCatalog(), mdl.getSchema(), RUNNING, Instant.now());
-            Task task = new Task(taskInfo, refreshCache(mdl));
-            tasks.put(taskId, task);
+            CatalogSchemaTableName catalogSchemaTableName = new CatalogSchemaTableName(mdl.getCatalog(), mdl.getSchema(), cacheInfo.getName());
+            TaskInfo taskInfo = new TaskInfo(mdl.getCatalog(), mdl.getSchema(), cacheInfo.getName(), RUNNING, Instant.now());
+            // To fix flaky test, we pass value to tasks instead of a reference;
+            Task task = new Task(TaskInfo.copyFrom(taskInfo), handleCache(mdl, cacheInfo));
```
brandboat, line 293: Could you explain more ? Not sure which test will be flaky without using copyFrom

accio-main/src/main/java/io/accio/main/web/CacheResource.java
```diff
@@ -63,14 +65,29 @@ public void reloadAsync(@Suspended AsyncResponse asyncResponse)
...
-                .getTaskInfo(taskId)
-                .thenApply(v -> v.orElseThrow(() -> new AccioException(NOT_FOUND, String.format("Task %s not found.", taskId))))
+                .getTaskInfo(catalogSchemaTableName)
+                .thenApply(v -> v.orElseThrow(() -> new AccioException(NOT_FOUND, String.format("Task %s not found.", catalogSchemaTableName))))
+                .whenComplete(bindAsyncResponse(asyncResponse));
+    }
+
+    @GET
+    @Path("info/{catalogName}/{schemaName}")
+    public void getTaskInfo(
```
brandboat, line 84: maybe getTaskInfos or getTaskInfoList

accio-tests/src/test/java/io/accio/testing/bigquery/TestReloadCache.java
```diff
@@ -95,22 +98,24 @@ public void testReloadCache()
...
         rewriteFile("cache/cache_reload_3_mdl.json");
         reloadAccioMDL();
-        taskInfo = reloadCacheAsync();
-        assertThat(taskInfo.getTaskStatus()).isEqualTo(RUNNING);
-        taskInfo = waitUntilFinished(taskInfo.getTaskId());
-        cachedTable = taskInfo.getCachedTables().get(0);
+        List<TaskInfo> taskInfos3 = reloadCacheAsync();
+        assertThat(taskInfos3.size()).isEqualTo(1);
+        taskInfo = taskInfos3.get(0);
+        // assertThat(taskInfo.getTaskStatus()).isEqualTo(RUNNING);
```
brandboat, line 112: Why do we comment out this line ?

accio-tests/src/test/java/io/accio/testing/bigquery/TestRefreshCache.java
```diff
@@ -41,14 +47,30 @@ protected Optional<String> getAccioMDLPath()
...
+        TaskInfo original = cacheManager.get().listTaskInfo(mdl.getCatalog(), mdl.getSchema()).join().stream()
+                .filter(taskInfo -> taskInfo.getCatalogSchemaTableName().equals(revenueName))
+                .findAny().orElseThrow(AssertionError::new);
+        Thread.sleep(6000);
+        // make sure the end time will be changed.
+        TaskInfo refreshed = getTaskInfoUntilDifferent(original);
+        assertThat(original.getEndTime()).isBefore(refreshed.getEndTime());
+    }
+
+    private TaskInfo getTaskInfoUntilDifferent(TaskInfo taskInfo)
```
brandboat, line 61: If we are using a active check loop with short sleep, then L55 (sleep 6s) could be removed.
And we should add a timeout for this method to avoid potential infinite while loop.

brandboat (approved):
LGTM % comments

accio-cache/src/main/java/io/accio/cache/CacheManager.java
liun03, line 293: A little confused: why`handleCache` calls `createTask`, and `createTask` also calls `handleCache`?

goldmedal: > A little confused: why`handleCache` calls `createTask`, and `createTask` also calls `handleCache`?

`createTask` means create a new cache. To get the latest end time, we need to create a new task periodically. That's why `handleCache` will use it in L127.

goldmedal: During testing, we will verify that the status of TaskInfo is set to 'RUNNING.' However, it's important to note that the task might conclude before relaying the information to the client. Therefore, we need to capture a snapshot at the task creation stage to accurately assess its status.

accio-tests/src/test/java/io/accio/testing/bigquery/TestReloadCache.java
goldmedal: Just forgot to change it back after fixing test.

accio-main/src/main/java/io/accio/main/web/CacheResource.java
```diff
@@ -66,7 +66,7 @@ public void reloadAsync(@Suspended AsyncResponse asyncResponse)
 
     @GET
     @Path("info/{catalogName}/{schemaName}/{tableName}")
-    public void getTaskInfo(
+    public void getTaskInfos(
```
brandboat, line 69: this only get one TaskInfo back
```suggestion
    public void getTaskInfo(
```

accio-tests/src/test/java/io/accio/testing/bigquery/TestRefreshCache.java
```diff
@@ -59,18 +65,24 @@ public void testRefreshFrequently()
...
-        while (refreshed == null ||
-                refreshed.getEndTime() == null ||
-                refreshed.getEndTime() == taskInfo.getEndTime()) {
-            refreshed = cacheManager.get().listTaskInfo(taskInfo.getCatalogName(), taskInfo.getSchemaName()).join().stream()
-                    .filter(t -> t.getTableName().equals(taskInfo.getTableName()))
-                    .findAny().orElseThrow(AssertionError::new);
-            Thread.sleep(100);
-        }
-        return refreshed;
+        return supplyAsync(() -> {
```
brandboat, line 70: Async is not required here. Could we use a timer ? And calculate `end-start` > `elapsed time`.
And if you think keep a long sleep here is required, maybe shorten it to `Thread.sleep(5000)` as we refresh cache every 5s. Add a comment like `// wait 5s for cache refresh`.

brandboat (approved):
2 nits

accio-cache/src/main/java/io/accio/cache/CacheManager.java
brandboat: as discussed offline, since TestRefreshCache#testRefreshSingleModel need to assert if TaskInfo status is `RUNNING`, that's why we use copyFrom here. To avoid task status was updated in other places.

accio-tests/src/test/java/io/accio/testing/bigquery/TestRefreshCache.java
goldmedal: The sleep shouldn't be equals to the refresh time. It must be less than the refresh time. If they're same, we could check the status when the refreshing is triggered every time.

brandboat: Sorry about making the confuse... I mean Thread.sleep(6000) in Line 61 , not the sleep in the while loop.

brandboat: BTW, MILLISECONDS.sleep(100) in while loop is reasonable for me.

goldmedal: oh.. seems the sleeping in L61 is unnecessary. Let's remove it.

PR #392, opened 2023-11-21
Add jgrapht to shaded

brandboat:
Since trino 375 use a very old version of jgraph (0.9.0) which is far away of what the version in accio (1.5.2). Upgrade jgrapht in trino 375 takes lots of effort. Here we add jgrapht to shaded jar.


goldmedal (approved):
LGTM

PR #398, opened 2023-12-04
Build multi arch image on main commit push

brandboat:
use docker buildx to build multi arch image, will build upon every commit push to main branch, and add image tags `main-{short_last_commit_sha}` and `latest`.


goldmedal (approved):
LGTM

PR #399, opened 2023-12-07
Redefine calculated and expression in model field

brandboat:
The "model expression" in the past allowed the mixing of column
names produced in refSql with relationship columns, causing
confusion for users. This PR redefines the usage scenarios for expr and calc.

- Currently, in the model, expr can only utilize columns from
the baseObject or those produced in refSql. Relationship columns
cannot be applied in expr. If one intends to use a relationship column,
it must be defined in calc. Additionally, calc can only utilize
columns within the model itself and cannot use columns from baseObject or refSql.
- And users should be aware that model primary key and relationship join key in
join critera shouldn't use `calc`.
- In metric, all expr in dimensions and measure are treated as calculated field.
- This pr also remove primary key and join key validation as they should be placed
in validation rules not in runtime.

### DIFF
--- a/accio-sqlrewrite/src/main/java/io/accio/sqlrewrite/ModelSqlRender.java
+++ b/accio-sqlrewrite/src/main/java/io/accio/sqlrewrite/ModelSqlRender.java
@@ -14,6 +14,7 @@
 
 package io.accio.sqlrewrite;
 
+import com.google.common.collect.ImmutableList;
 import io.accio.base.AccioMDL;
 import io.accio.base.dto.Column;
 import io.accio.base.dto.Model;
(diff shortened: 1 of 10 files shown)


accio-sqlrewrite/src/main/java/io/accio/sqlrewrite/ModelSqlRender.java
```diff
@@ -147,35 +116,115 @@ private void processRelationshipInfos(Column column, Model baseModel, List<Expre
...
     @Override
-    protected String getCalculatedSubQuery(Model baseModel, CalculatedFieldRelationshipInfo relationshipInfo)
+    protected List<SubQueryJoinInfo> getCalculatedSubQuery(Model baseModel, List<CalculatedFieldRelationshipInfo> relationshipInfos)
     {
-        String requiredExpressions = format("%s AS \"%s\"",
-                RelationshipRewriter.rewrite(relationshipInfo.getExpressionRelationshipInfo(), parseExpression(relationshipInfo.getColumn().getExpression().get())),
-                relationshipInfo.getAlias());
+        ImmutableList.Builder<SubQueryJoinInfo> queries = ImmutableList.builder();
+        queries.addAll(getToOneRelationshipsQuery(baseModel, relationshipInfos));
+        queries.addAll(getToManyRelationshipsQuery(baseModel, relationshipInfos));
```
goldmedal, line 138: Consider rename `getToOneRelationshipsQuery` and `getToManyRelationshipsQuery`.
We distinguish them according to doing aggregation or not. Maybe rename them to
- `getMapExpressionQuery`: allow only to_one relationship
- `getReduceExpressionQuery`: allow both to_one and to_many
Inspire form the concept of `map` and `reduce`.

goldmedal (approved):
Minor commets. Others LGTM

accio-sqlrewrite/src/main/java/io/accio/sqlrewrite/ModelSqlRender.java
brandboat: I still prefer the original method name, will add more comment to methods.

PR #412, opened 2023-12-25
Add DuckDBConfig to handle the system resource configurations

goldmedal:
# Description
There're 3 configurations exposed by this PR:
- `duckdb.memory-limit`: To set `memory_limit` in DuckDB. The default value is `the runtime max heap size / 2`
- `duckdb.home-directory`: To set `home-directory` in DuckDB.
  - This config is disable because the known issue: https://github.com/duckdb/duckdb/issues/10062
-  `duckdb.temp-directory`: To set `temp-driectory` in DuckDB.

### DIFF
--- a/accio-base/src/main/java/io/accio/base/client/duckdb/DuckDBConfig.java
+++ b/accio-base/src/main/java/io/accio/base/client/duckdb/DuckDBConfig.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package io.accio.base.client.duckdb;
+
+import io.airlift.configuration.Config;
+
+public class DuckDBConfig
+{
+    private String memoryLimit = Runtime.getRuntime().maxMemory() / 2 + "B";
+    private String homeDirectory = "/tmp/duckdb";
+    private String tempDirectory = "/tmp/duck";
+
+    public String getMemoryLimit()
+    {
+        return memoryLimit;
+    }
+
+    @Config("duckdb.memory-limit")
+    public void setMemoryLimit(String memoryLimit)
+    {
+        this.memoryLimit = memoryLimit;
+    }
+
+    public String getHomeDirectory()
+    {
+        return homeDirectory;
+    }
+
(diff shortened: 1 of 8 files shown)


AShiou (approved):

accio-base/src/main/java/io/accio/base/client/duckdb/DuckDBConfig.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+package io.accio.base.client.duckdb;
+
+import io.airlift.configuration.Config;
+
+public class DuckDBConfig
+{
+    private String memoryLimit = Runtime.getRuntime().maxMemory() / 2 + "B";
+    private String homeDirectory = "/tmp/duckdb";
```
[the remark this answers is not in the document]
goldmedal: hmm.. or we don't give the default value for it? It's hard to decide where to place it.

accio-main/src/main/java/io/accio/main/web/CacheResource.java
```diff
@@ -99,7 +99,7 @@ public void getDuckDBSettings(@Suspended AsyncResponse asyncResponse)
             asyncResponse.resume(Response.ok(cacheManager.getDuckDBSettings()).build());
         }
         catch (Exception e) {
-            asyncResponse.resume(Response.serverError().build());
+            asyncResponse.resume(asyncResponse.resume(e));
```
brandboat, line 102: one last comment, perhaps we can omit the try-catch block in this section. It seems that the AccioExceptionMapper is set up to catch the error and return a 500 response in its toResponse method. However, I'm not entirely certain about this, could you help me check this?

goldmedal: Yes, If we throw the runtime exception from `cacheManager.getDuckDBSettings`, we can omit the try-catch.

brandboat (approved):
LGTM

grieve54706:
@goldmedal Do we check the path if it does not exist and the feature still works? I tried the home_directory with a non-exist will not work. However, using an existing path will work in the session.

I also check your PR in the DuckDB about `home_directory`. I tried the way `SET home_directory; SELECT current_setting('home_directory');` in one statement and it got the value. But I tried another way `SET home_directory;` and `SELECT current_setting('home_directory');` with different connections will not work.

goldmedal:
> @goldmedal Do we check the path if it does not exist and the feature still works? I tried the home_directory with a non-exist will not work. However, using an existing path will work in the session.
> 
I think we don't need to check it. We just pass the config to DuckDB. This config should checked by DuckDB. 

> I also check your PR in the DuckDB about `home_directory`. I tried the way `SET home_directory; SELECT current_setting('home_directory');` in one statement and it got the value. But I tried another way `SET home_directory;` and `SELECT current_setting('home_directory');` with different connections will not work.

I has checked this issue is a bug or behavior of DuckDB. https://github.com/duckdb/duckdb/issues/10062
That's why I disable the feature temporarlly.
https://github.com/Canner/accio/blob/33442e032219eb35da44ae1454a6c614a5d89768/accio-base/src/main/java/io/accio/base/client/duckdb/DuckdbClient.java#L75-L77

PR #419, opened 2024-01-04
Fix prepared statement parameter for BIgQuery

goldmedal:
- Support more type parameter passed by native protocol.
  - `numeric`, `date`, `timestamp`, `timestamptz`, `bytea`, `interval`, `array`.

### DIFF
--- a/accio-base/src/main/java/io/accio/base/type/TimestampWithTimeZoneType.java
+++ b/accio-base/src/main/java/io/accio/base/type/TimestampWithTimeZoneType.java
@@ -31,7 +31,6 @@ import java.util.Locale;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.time.format.DateTimeFormatter.ISO_LOCAL_DATE;
 import static java.time.format.DateTimeFormatter.ISO_LOCAL_TIME;
-import static java.time.format.DateTimeFormatter.ISO_ZONED_DATE_TIME;
 import static java.time.temporal.ChronoField.MILLI_OF_SECOND;
 
 public class TimestampWithTimeZoneType
@@ -89,14 +88,8 @@ public class TimestampWithTimeZoneType
     @VisibleForTesting
     ZonedDateTime tryParse(String timeString)
     {
-        try {
-            // Postgres TimestampTz format
-            return ZonedDateTime.parse(timeString, PG_TIMESTAMP);
-        }
-        catch (Exception e) {
-            // JDBC transform ZonedDateTime to String
-            return ZonedDateTime.parse(timeString, ISO_ZONED_DATE_TIME);
-        }
+        // Postgres TimestampTz format
+        return ZonedDateTime.parse(timeString, PG_TIMESTAMP);
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


accio-connector-client/src/main/java/io/accio/connector/bigquery/BigQueryType.java
```diff
@@ -95,6 +96,7 @@ private BigQueryType() {}
                 .put(INTEGER, INT64)
                 .put(SMALLINT, INT64)
                 .put(TINYINT, INT64)
+                .put(OidType.OID_INSTANCE, INT64)
```
brandboat, line 99: nit: static import
```suggestion
                .put(OID_INSTANCE, INT64)
```

accio-tests/src/main/java/io/accio/testing/TestingWireProtocolClient.java
```diff
@@ -1129,6 +1131,22 @@ public FormatCodes.FormatCode getFormatCode()
...
+                            stringBuilder.append("NULL");
+                        }
+                        else {
+                            stringBuilder.append(element);
+                        }
+                        stringBuilder.append(',');
+                    }
+                    stringBuilder.setLength(stringBuilder.length() - 1);
+                    stringBuilder.append("}");
+                    return stringBuilder.toString().getBytes(UTF_8);
```
brandboat, line 1148: Pardon, is this another pg jdbc behavior or something that was defined in pg wire protocol ? Not sure why the PGArray format is `{object,object,object}`

accio-base/src/main/java/io/accio/base/type/TimestampWithTimeZoneType.java
```diff
@@ -88,12 +89,25 @@ public Object decodeUTF8Text(byte[] bytes)
...
     {
-        return ZonedDateTime.parse(timeString, PG_TIMESTAMP);
+        try {
+            // Postgres TimestampTz format
+            return ZonedDateTime.parse(timeString, PG_TIMESTAMP);
+        }
+        catch (Exception e) {
+            // JDBC transform ZonedDateTime to String
+            return ZonedDateTime.parse(timeString, ISO_ZONED_DATE_TIME);
+        }
```
brandboat, line 99: Pardon, this makes me a little bit confuse, why do we need to handle two different formats here ?

brandboat (approved):
some question, overall lgtm

accio-tests/src/main/java/io/accio/testing/TestingWireProtocolClient.java
goldmedal: https://github.com/Canner/accio/blob/80e3bcfb5c7a9843aaed128564782f18ceff6749/accio-base/src/main/java/io/accio/base/type/PGArray.java#L280-L283

It's the definition in PG Wire Protocol.

accio-base/src/main/java/io/accio/base/type/TimestampWithTimeZoneType.java
goldmedal: oh, it's my mistake. I should use `PG_TIMESTAMP` to format `ZoneDateTime` in the test. We don't need to handle `ISO_ZONED_DATE_TIME` format.

PR #420, opened 2024-01-05
Fix LIMIT OFFSET for BigQuery DIalect

goldmedal:


trino-parser/src/main/java/io/trino/sql/SqlFormatter.java
```diff
@@ -378,8 +379,14 @@ protected Void visitQuery(Query node, Integer indent)
...
-            node.getOffset().ifPresent(offset -> process(offset, indent));
-            node.getLimit().ifPresent(limit -> process(limit, indent));
+            if (dialect.equals(BIGQUERY)) {
+                node.getLimit().ifPresent(limit -> process(limit, indent));
+                node.getOffset().ifPresent(offset -> process(offset, indent));
+            }
+            else {
+                node.getOffset().ifPresent(offset -> process(offset, indent));
+                node.getLimit().ifPresent(limit -> process(limit, indent));
+            }
```
brandboat, line 389: Does the order matters ? Looks like both if-else blocks are almost doing the same thing.

trino-parser/src/main/java/io/trino/sql/SqlFormatter.java
```diff
@@ -414,8 +421,14 @@ protected Void visitQuerySpecification(QuerySpecification node, Integer indent)
...
```
brandboat, line 431: Same thing here.

accio-main/src/main/java/io/accio/main/connector/bigquery/BigQuerySqlConverter.java
```diff
@@ -81,11 +81,14 @@ public String convert(@Language("sql") String sql, SessionContext sessionContext
         LOG.info("[Input sql]: %s", sql);
 
         for (SqlRewrite rewrite : sqlRewrites) {
-            LOG.debug("Before %s: %s", rewrite.getClass().getSimpleName(), formatSql(rewrittenNode));
+            if (LOG.isDebugEnabled()) {
```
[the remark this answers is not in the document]
goldmedal: IMO, If we don't add this condition, `formatSql` will be invoked but won't be show.

trino-parser/src/main/java/io/trino/sql/SqlFormatter.java
goldmedal: Yes, it's an ordering problem. BigQuery need to place `OFFSET` behind `LIMIT`.

brandboat (approved):
lgtm

PR #422, opened 2024-01-09
Take care cumulative metric in enable dynamic fields mode

brandboat:
note that even enable dynamic fields, every field in cumulative metric are still evolved since cumulative metric is a static sql pattern, and itself couldn't be dynamic.


brandboat:
already rebased to the latest commit

goldmedal (approved):
LGTM

PR #429, opened 2024-01-12
Pass AnalyzedMDL to sqlrewrite

brandboat:
pass AnalyzedMDL to AccioSqlRewrite to avoid run data lineage in every sql.


goldmedal (approved):
LGTM

PR #433, opened 2024-01-16
Add properties to all DTOs and use toStringHelper

brandboat:
also mark description `@Deprecated`

### DIFF
--- a/accio-base/src/main/java/io/accio/base/dto/Measure.java
+++ b/accio-base/src/main/java/io/accio/base/dto/Measure.java
@@ -17,31 +17,35 @@ package io.accio.base.dto;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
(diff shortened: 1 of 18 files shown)


accio-base/src/main/java/io/accio/base/dto/Measure.java
```diff
@@ -78,15 +88,16 @@ public boolean equals(Object o)
...
-        return Objects.equals(name, measure.name) &&
-                Objects.equals(type, measure.type) &&
-                Objects.equals(operator, measure.operator) &&
-                Objects.equals(refColumn, measure.refColumn);
+        return Objects.equals(name, measure.name)
+                && Objects.equals(type, measure.type)
+                && Objects.equals(operator, measure.operator)
+                && Objects.equals(refColumn, measure.refColumn)
+                && Objects.equals(properties, measure.properties);
     }
```
goldmedal, line 97: Measure doesn't override `toString`? Could you add it?

PR #439, opened 2024-01-23
Add analysis sql web api

brandboat:
# Description
Add new web api /v1/analysis/api, this api will retrieve predicates in where-clause. Note that this api only catch predicate format in `[Identifer/dereferenceExpression] [operator] [literal/identifier/dereferenceExpression]`. e.g. `column = 'foo'` will be caught while `column_a = column_b` won't be caught since column_b is an identifier not literal.

## input dto
manifest is optional, if manifest is not provided in input dto, mdl provided in accio server will be used.
```json
{
  "manifest":  {...},
  "sql": "SELECT * FROM Orders WHERE orderdate >= DATE '2022/07/12' AND orderkey <= 123 OR orderkey >= 100"
}
```

## output dto
```json
[
  {
     "modelName": "Orders",
     "columnName": "orderdate",
     "predicates": [
       {
         "operator": ">=",
         "value": "DATE '2022/07/12'"
       }
     ]
  },
  {
     "modelName": "Orders",
     "columnName": "orderkey"
     "predicates": [
       {
         "operator": "<=",
         "value": "123"
       },
       {
         "operator": ">=",
         "value": "100"
       }
     ]
  }
]
```

### DIFF
--- a/accio-base/src/main/java/io/accio/base/sqlrewrite/analyzer/matcher/PredicateMatcher.java
+++ b/accio-base/src/main/java/io/accio/base/sqlrewrite/analyzer/matcher/PredicateMatcher.java
@@ -1,0 +1,29 @@
+package io.accio.base.sqlrewrite.analyzer.matcher;
+
+import io.trino.sql.tree.ComparisonExpression;
+import io.trino.sql.tree.DereferenceExpression;
+import io.trino.sql.tree.Expression;
+import io.trino.sql.tree.Identifier;
+import io.trino.sql.tree.Literal;
+import io.trino.sql.tree.Node;
+
+public class PredicateMatcher
+        implements Matcher
+{
+    public static final PredicateMatcher PREDICATE_MATCHER = new PredicateMatcher();
+
+    private PredicateMatcher() {}
+
+    @Override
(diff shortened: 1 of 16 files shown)


accio-base/src/main/java/io/accio/base/sqlrewrite/analyzer/matcher/PredicateMatcher.java
```diff
@@ -0,0 +1,29 @@
...
+    public boolean shapeMatches(Node node)
+    {
+        if (!(node instanceof ComparisonExpression)) {
+            return false;
+        }
+
+        Expression left = ((ComparisonExpression) node).getLeft();
+        Expression right = ((ComparisonExpression) node).getRight();
+
+        return (left instanceof DereferenceExpression || left instanceof Identifier) && right instanceof Literal;
```
goldmedal, line 27: Should we match the pattern with another column at the right side? Some SQL like
```sql
select * from tableA where time1 > time2
```

goldmedal (approved):
LGTM. Some minor comment and question.

goldmedal (approved):
LGTM

PR #440, opened 2024-01-24
Shutdown all executors to avoid thread leak

brandboat:
This won't happen in production since the only way to close accio server means kill process. 
While some executor is not shutdown and cause thread leak in e2e tests.


goldmedal (approved):
Thanks. LGTM.

PR #442, opened 2024-01-25
Implement basic Type Coercion for BigQuery connector

goldmedal:
# Description
Refer to [BigQuery Type Conversion Rule](https://cloud.google.com/bigquery/docs/reference/standard-sql/conversion_rules), it's very weak to do timestamp-related type conversion. If the timestamp-related type can't be work with other time type, such as `DATE`, `TIME`, it's very inconvenience for normal SQL user.

In this PR, we supports to detect the Type Coercion request automatically. Any type coercion request, we may wrap what need to be coerced in a `CAST` expression. For now, any type coercion is based on what type left side is.

Now, user can query like as below, assume `shipdate` is a `TIMESTAMP` column in BigQuery.
```
SELECT shipdate FROM Lineitem WHERE shipdate > current_date + interval '1' day
```

# TODO Issue
- Function list for return type.
- Handle remote function name.
- Handle timestamp with tz type
- support to analyze quantified comparison type
- support bq datetime type (handle type not int pg)
- more pg type  name alias

### DIFF
--- a/accio-base/src/main/java/io/accio/base/sqlrewrite/analyzer/StatementAnalyzer.java
+++ b/accio-base/src/main/java/io/accio/base/sqlrewrite/analyzer/StatementAnalyzer.java
@@ -47,6 +47,8 @@ import io.trino.sql.tree.Values;
 import io.trino.sql.tree.With;
 import io.trino.sql.tree.WithQuery;
 
+import javax.annotation.Nullable;
+
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -72,8 +74,14 @@ public final class StatementAnalyzer
 
     public static Analysis analyze(Statement statement, SessionContext sessionContext, AccioMDL accioMDL)
     {
+        return analyze(statement, sessionContext, accioMDL, null);
+    }
+
+    public static Analysis analyze(Statement statement, SessionContext sessionContext, AccioMDL accioMDL, TypeCoercion typeCoercion)
+    {
         Analysis analysis = new Analysis(statement);
-        new Visitor(sessionContext, analysis, accioMDL).process(statement, Optional.empty());
+        Scope queryScope = new Visitor(sessionContext, analysis, accioMDL, typeCoercion).process(statement, Optional.empty());
+        analysis.setQueryScope(queryScope);
 
         // add models directly used in sql query
         analysis.addModels(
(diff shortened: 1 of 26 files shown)


brandboat:
Could you help create a umbrella issue for this ?
- Function list for return type.
- Handle remote function name.
- Handle timestamp with tz type
- support to analyze quantified comparison type
- support bq datetime type (handle type not int pg)
- more pg type name alias

accio-base/src/main/java/io/accio/base/sqlrewrite/analyzer/StatementAnalyzer.java
```diff
@@ -150,55 +161,25 @@ protected Scope visitTable(Table node, Optional<Scope> scope)
...
-                                    .name(cumulativeMetric.getMeasure().getName())
-                                    .build());
+                List<Field> fields = collectFieldFromMDL(tableName);
+
+                // if catalog and schema matches, but table name doesn't match any model, we assume it's a remote data source table
+                if (fields.isEmpty()) {
+                    return Scope.builder()
+                            .parent(scope)
+                            .isDataSourceScope(true)
+                            .build();
```
brandboat, line 182: If I remember correctly, we don't allow user to directly select remote data source table, not sure why we need this.

brandboat (approved):
LGTM % questions

accio-base/src/main/java/io/accio/base/sqlrewrite/analyzer/StatementAnalyzer.java
goldmedal: To analyze the type of an identifier or a dereference expression, we need to get the correct metadata from models. Refer to the pattern to generate the model query:
```sql
WITH Orders as (
SELECT
  o_orderkey as orderkey
  ...
FROM
   (
     SELECT * FROM bq.tpch.orders
   ) Orders
)
```
The query for the remote data source is always wrapped in an `AliasedRelation` and the alias name will be the model name. The scoping fields of this subquery is always empty because we can't get the column metadata from the remote data source. Thus, we give a tag `isDataSourceScope` for this kind of relation, then get the column metadata based on the alias name(model name) from MDL to build the correct scope.

brandboat: Thanks for explanation, I got mislead by the comment `we assume it's a remote data source table`, actually we still get metadata from mdl, not remote data source.

PR #443, opened 2024-01-26
Test single threaded for mvn test

grieve54706:
We found the ci `integration-tests` will stuck after `testStatementNameWithHyphens`.
And we found the sign when the method done, the process will not trigger `@AfterMethod`. We guess the maven could be deranged when multi-thread.


grieve54706:
We don't find any flaky.

PR #450, opened 2024-01-30
Add CI to build image with specify branch and commit

grieve54706:
## Description
The new workflow can receive a specific branch to build the image. It also can receive a specific docker image tag name instead of the default name `branchName-commitSHA`.

image

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -3,9 +3,9 @@ name: Build image
 on:
   workflow_dispatch:
     inputs:
-      docker_image_name:
+      docker_image_tag_name:
         type: string
-        description: Docker image name (Optional)
+        description: Docker image tag name (Optional)
 
 jobs:
   build-image:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-image.yml
```diff
@@ -0,0 +1,58 @@
...
+        type: string
+        description: Specify the commit hash (Optional)
+
+jobs:
+  build-image:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.specify_commit || github.ref }}
```
brandboat, line 19: Pardon me, if I enter `image_tag_name` will this checkout to the given tag branch ?

goldmedal: Look like `image_tag_name` is only a prefix for tag name (?
@grieve54706  Is it possible to select a branch name and utilize its latest commit?

brandboat (approved):
lgtm, thanks

grieve54706:
After discussing, we don't need specific commit. I will remove it. And short the description of the docker image name.

.github/workflows/build-image.yml
grieve54706: I modified the field name to `docker_image_name` and removed the field `specify_commit`.
The `docker_image_name` is only for instead of the default docker image name.

grieve54706: I think the `docker_image_name` seem not appropriate. I will modify it to `docker_image_tag_name`.

goldmedal (approved):
lgtm thanks

PR #451, opened 2024-01-31
Create dummy CTE when selecting table without any required field in dynamic mode

goldmedal:
# Description
Some BI tools will select a table without any required field like `Metabase` query a table using:
```sql
SELECT true "_"
FROM
  Metric_Revenue
WHERE (1 <> 1)
LIMIT 0
```

We should create a dummy CTE when required fields is empty but exist selected tables.


brandboat (approved):
lgtm, thanks

PR #453, opened 2024-02-01
Fix the scope analyzer for with query statement

goldmedal:
# Description
Fix the query scope analysis for the WITH query. The wrong analysis will make the type analysis and type coercion be wrong.


grieve54706 (approved):
LGTM

PR #454, opened 2024-02-02
Fix collect required fields in where statement in dynamic calculated mode

goldmedal:
We missed to collect the columns when analyze where statement.


grieve54706 (approved):
LGTM

PR #455, opened 2024-02-03
Add quote for all identifier to support SQL reserved word as name

goldmedal:
Add quote for all column name or table name identifier to support reserved SQL reserved word as name.

# Known issue
- The test in 17595f9 will be failed in CI but work in the local.


grieve54706 (approved):
LGTM

PR #457, opened 2024-02-06
Throw error if a specific table isn't found

goldmedal:
After #451, we will create dummy CTE for required table but we should exclude the table which doesn't exist actually.


grieve54706 (approved):
LGTM

Issue #461, opened 2024-02-16
Support psycopg2 driver


goldmedal:
`psycopg2` is a popular PG driver in python. Sample connection code:
```python
import psycopg2

conn = psycopg2.connect(
    dsn='postgres://root:ff@localhost:7432/canner-cml?options=--search_path%3Dhubspot'
)

with conn.cursor() as cur:
    sql = "select count(*) from Model_Customer"
    cur.execute(sql)
```

PR #463, opened 2024-02-16
Support psycopg2 via ignoring session properties

grieve54706:
## Description
`psycopg2` will send set-statement with `datestyle`.
We need to ignore the property.

## Related issues
#461


goldmedal (approved):
LGTM

PR #466, opened 2024-02-20
Add function registry bundle

### LINKED ISSUE #444: Create a supported function list
As the mentioned in #442, we should create a function list to handle all function we supported. It will be used to check the return type when analyze the type coercion request.

grieve54706:
## Description
In the future, we will add many registries for different data sources. We need a bundle for them. The ExpressionTypeAnalyzer can get the function via the bundle.

## Related issues
Resolve #444


grieve54706:
Due to restructuring on the main branch, a new PR will be resubmitted

PR #470, opened 2024-02-22
Support function overloading

### LINKED ISSUE #92: Enhance function overloading support
FunctionKey only recognizes each function by its name and number of argument now. We should try to analyze the type of each argument, then we can support the function overloading with same number of argument but different types.

grieve54706:
## Description
Resolve #92.
Use argument types instead of argument count.
I found some issues about scope, so I commented out them and created issues to follow them.


goldmedal:
We have other plan to handle the function list. Let's close this temporarily.

PR #471, opened 2024-02-23
Enhance the scope analysis in StatementAnalyzer

goldmedal:
# Description
To support the more complex SQL analysis, we should enhance the SQL analysis firstly.
# Changed
- Collect all scope info for each node to a map. It's useful for generating the rewrite plan in the future.
- Assign the relation Id for every scope.


grieve54706 (approved):
LGTM

PR #475, opened 2024-02-29
Upgrade azul/zulu-openjdk from 11.0.14 to 11.0.22 for arm64

grieve54706:
azul/zulu-openjdk:11.0.14 only has amd64, but 11.0.22 has `arm64`.


goldmedal (approved):
LGTM

PR #478, opened 2024-03-04
Fix for PostgreSQL as data source

grieve54706:
We tried to make PostgreSQL as data source and use Metabase to connect Accio. We found some issues.

PgMetastore should bind DuckDBMetadata not PostgresMetadata. Although the Data source is PostgreSQL, the metastore still uses DuckDB in our design.

Metabase will send a statement like `select * from "orders"` to Accio. But Accio remove the quoted when the rewriter rewrite. We should not change the quoted delimited status.


goldmedal (approved):
LGTM

PR #480, opened 2024-03-06
Merge accio-connector-client module to accio-main

goldmedal:
`accio-connector-client` only used by `accio-main`. We don't need to split them to different modules.


grieve54706 (approved):
LGTM

PR #481, opened 2024-03-08
Implement ConfigManager for modified configurations in runtime

goldmedal:
# Description
In this PR, we implemented `ConfigManger` to manage all configurations. All config will be got from a centralize manager which maybe changed in runtime.

# New API for configuration change
## Rest all config
This API will reset all configuration to default value.
```
DELETE /v1/config
```
### Response
```
HTTTP 200
```

## Update Config
This API only updates the specific configs. Everytime update configs, server will archived the old one to to a folder created in configs file path called `archived`.
```
PATCH /v1/config
```
### Input format
```
[
	{
		"name": "",
		"value": ""
	},
	{
		"name": "",
		"value": ""
	}
]
```
### Response
```
HTTP 200
```

## List All config
Get all the config
```
GET /v1/config
```
### output format
```
HTTP 200
[
	{
		"name": "",
		"value": ""
	},
	{
		"name": "",
		"value": ""
	}
	...
]
```

## Get one config
```
GET /v1/conf/{configName}
```

### Output format
```
HTTP 200
	{
		"name": "",
		"value": ""
	}
```

# Some Configuration should never be changed in runtime
```
pg-wire-protocol.ssl.enabled
pg-wire-protocol.netty.thread.count
pg-wire-protocol.auth.file
pg-wire-protocol.port

@deprecated accio.file
accio.directory
accio.datasource.type
```

# Known Issues
- For now, the configuration change in runtime won't be applied to the data source connector. This issue will be resolved in next PR.


accio-main/src/main/java/io/accio/main/web/ConfigResource.java
```diff
@@ -0,0 +1,91 @@
...
+            List<ConfigManager.ConfigEntry> configEntries,
+            @Suspended AsyncResponse asyncResponse)
+    {
+        CompletableFuture
+                .runAsync(() -> {
+                    configManager.reset();
+                    configManager.setConfigs(configEntries);
+                })
+                .whenComplete(bindAsyncResponse(asyncResponse));
+    }
```
grieve54706, line 79: Could you help me know what scenario needs this api?

accio-base/src/main/java/io/accio/base/config/ConfigManager.java
```diff
@@ -0,0 +1,345 @@
...
+            });
+        }
+        throw new RuntimeException("Unknown config class: " + config.getName());
+    }
+
+    public synchronized void setConfigs(List<ConfigEntry> configEntries)
+    {
+        configEntries.forEach(entry -> setConfigInternal(entry.getName(), entry.getValue()));
+        resetCache();
+    }
```
grieve54706, line 237: Should we update the config file?

accio-base/src/main/java/io/accio/base/config/ConfigManager.java
```diff
@@ -0,0 +1,345 @@
...
+public class ConfigManager
+{
+    private Optional<AccioConfig> accioConfig;
+    private Optional<PostgresConfig> postgresConfig;
+    private Optional<BigQueryConfig> bigQueryConfig;
+    private Optional<DuckDBConfig> duckDBConfig;
+    private Optional<PostgresWireProtocolConfig> postgresWireProtocolConfig;
+    private Optional<DuckdbS3StyleStorageConfig> duckdbS3StyleStorageConfig;
+
+    private ConcurrentHashMap<String, String> configs = new ConcurrentHashMap<>();
```
grieve54706, line 77: I thought we will just update the PostgresConfig and BigQueryConfig, etc. Why we specially save the config cache?

accio-server/src/main/java/io/accio/server/module/ConnectorConfigModule.java
```diff
@@ -0,0 +1,33 @@
...
+package io.accio.server.module;
+
+import com.google.inject.Binder;
+import io.accio.base.config.BigQueryConfig;
+import io.accio.base.config.PostgresConfig;
+import io.airlift.configuration.AbstractConfigurationAwareModule;
+
+import static io.airlift.configuration.ConfigBinder.configBinder;
+
+public class ConnectorConfigModule
```
grieve54706, line 24: Why we need to pull up the config to this module? I believe they should be close who use. And some config do not need to bind when the data source type is not with them.

accio-tests/src/test/java/io/accio/testing/TestConfigResource.java
```diff
@@ -0,0 +1,93 @@
...
+            Path accioMDLFilePath = mdlDir.resolve("acciomdl.json");
+            Files.write(accioMDLFilePath, MANIFEST_JSON_CODEC.toJsonBytes(manifest));
+        }
+        catch (IOException ex) {
+            throw new RuntimeException(ex);
+        }
+
+        ImmutableMap.Builder<String, String> properties = ImmutableMap.<String, String>builder()
+                .put(ACCIO_DIRECTORY, mdlDir.toAbsolutePath().toString())
+                .put(ACCIO_DATASOURCE_TYPE, "duckdb");
```
grieve54706, line 57: We can use `DataSourceType.DUCKDB.name()` instead of the string to reduce string. WDYT?

accio-tests/src/test/java/io/accio/testing/bigquery/TestBigQueryCacheService.java
```diff
@@ -72,12 +75,18 @@ private TestBigQueryCacheService()
                 .setCredentialsKey(getenv("TEST_BIG_QUERY_CREDENTIALS_BASE64_JSON"))
                 .setProjectId(getenv("TEST_BIG_QUERY_PROJECT_ID"))
                 .setBucketName(getenv("TEST_BIG_QUERY_BUCKET_NAME"));
-        BigQueryCredentialsSupplier bigQueryCredentialsSupplier = provideBigQueryCredentialsSupplier(bigQueryConfig);
-        HeaderProvider headerProvider = createHeaderProvider();
-        BigQueryClient bigQueryClient = provideBigQuery(bigQueryConfig, headerProvider, bigQueryCredentialsSupplier);
+        BigQueryCredentialsSupplier bigQueryCredentialsSupplier = new BigQueryCredentialsSupplier(bigQueryConfig.getCredentialsKey(), bigQueryConfig.getCredentialsFile());
+        HeaderProvider headerProvider = FixedHeaderProvider.create("user-agent", "accio/1");
+        ConfigManager configManager = new ConfigManager(
```
grieve54706, line 80: Could we add TestingConfigManager with builder pattern to reduce useless config?

accio-base/src/main/java/io/accio/base/config/ConfigManager.java
```diff
@@ -0,0 +1,345 @@
...
+        if (value == null) {
+            // ConcurrentHashMap are not allowed value to be null, use empty string instead.
+            configs.put(key, "");
+        }
+        else {
+            configs.put(key, value);
+        }
+    }
+
+    public <T> T getConfig(Class<T> config)
```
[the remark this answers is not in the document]
goldmedal: I preferred this way to get the specific configuration. It's more flexible.

accio-tests/src/test/java/io/accio/TestBigQuerySqlConverter.java
```diff
@@ -50,12 +54,15 @@ public void createBigQueryClient()
...
-        BigQueryClient bigQueryClient = BigQueryConnectorModule.provideBigQuery(
+        ConfigManager configManager = new ConfigManager(
+                new AccioConfig(),
+                new PostgresConfig(),
                 config,
-                BigQueryConnectorModule.createHeaderProvider(),
-                BigQueryConnectorModule.provideBigQueryCredentialsSupplier(config));
+                new DuckDBConfig(),
+                new PostgresWireProtocolConfig(),
+                new DuckdbS3StyleStorageConfig());
```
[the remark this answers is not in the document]
goldmedal: This ctor is only used here. I think it's not worth to write builder pattern for it.

accio-base/src/main/java/io/accio/base/config/ConfigManager.java
goldmedal: no, we only handle all configs in the memory layer.

accio-main/src/main/java/io/accio/main/web/ConfigResource.java
goldmedal: If user wants to clear all configurations and set a new one, he can use this.

accio-base/src/main/java/io/accio/base/config/ConfigManager.java
goldmedal: Using Map to handle all configs, it can help us to manage the configurations efficiently. By cache all config, we don't need to create a new config instance when we want to get the latest configuration everytime.

accio-tests/src/test/java/io/accio/testing/bigquery/TestBigQueryCacheService.java
goldmedal: I'm not sure what you mean but I think we should assert other unused configurations won't be impacted when we only set the specific configuration.

accio-server/src/main/java/io/accio/server/module/ConnectorConfigModule.java
goldmedal: When ConfigManager be created, all of them should be injected.

accio-tests/src/test/java/io/accio/testing/TestConfigResource.java
goldmedal: hmm. I think we can keep this to assert this configuration is case insensitive.

accio-base/src/main/java/io/accio/base/config/ConfigManager.java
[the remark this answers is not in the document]
grieve54706: You can keep this method and add other method for some scenario explicitly get which config.

accio-tests/src/test/java/io/accio/testing/bigquery/TestBigQueryCacheService.java
grieve54706: They should be not impacted each other. We don't over design them.

accio-main/src/main/java/io/accio/main/web/ConfigResource.java
goldmedal: As the offline discussion in last time, the configuration file will be maintained by web side.

accio-tests/src/test/java/io/accio/testing/TestConfigResource.java
grieve54706: Case insensitive should be other test case. Not append with this case.

accio-server/src/main/java/io/accio/server/module/ConnectorConfigModule.java
grieve54706: Not make sense. If the scenario is data source type is PG, why we should assert BQ config be injected?

accio-base/src/main/java/io/accio/base/config/ConfigManager.java
grieve54706: The runtime config is different with config file is weired.

accio-base/src/main/java/io/accio/base/config/ConfigManager.java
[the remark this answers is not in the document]
goldmedal: As the offline discussion, we split it by different config to different method to avoid a function too big.

accio-main/src/main/java/io/accio/main/web/ConfigResource.java
goldmedal: As the discussion, we remove the update api and implement reset instead.

accio-base/src/main/java/io/accio/base/config/ConfigManager.java
goldmedal: As the discussion, we will sync memory to the config file and archived the original one.

accio-server/src/main/java/io/accio/server/module/ConnectorConfigModule.java
goldmedal: For the current architecture, we load all config when server start but switch them in runtime.

grieve54706 (approved):
LGTM

PR #482, opened 2024-03-11
Implement Dynamic DataSource connector

goldmedal:
follow up #481 
# Description
This PR allow user to change the data source config after server started. Every time to change configs, system will detect if this change needs to reload the connector. 
We define 3 kind of configuration:
- Static Config: The configuration can't be changed. Only set when server start.
- Reload Required Config: The configuration can be changed and will trigger the reload.
- Normal Config: The configuration can be changed. No need to reload.

# Static Config
```
accio.file
accio.directory
duckdb.storage.endpoint
pg-wire-protocol.port
pg-wire-protocol.netty.thread.count
pg-wire-protocol.auth.file
```

# Reload Required Reload:
```
accio.datasource.type
accio.experimental-enable-dynamic-fields
duckdb.storage.access-key
duckdb.storage.secret-key"
duckdb.storage.region
duckdb.memory-limit
duckdb.home-directory
duckdb.temp-directory
duckdb.max-concurrent-tasks
duckdb.max-concurrent-queries
duckdb.max-cache-query-timeout
duckdb.max-cache-table-size-ratio
duckdb.cache-task-retry-delay
bigquery.credentials-key
bigquery.project-id
bigquery.location
bigquery.bucket-name
bigquery.metadata.schema.prefix
postgres.jdbc.url
postgres.user
postgres.password
```


accio-main/src/main/java/io/accio/main/sql/SqlConverterManager.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @Inject
+    public SqlConverterManager(
+            ConfigManager configManager,
+            BigQuerySqlConverter bigQuerySqlConverter,
+            PostgresSqlConverter postgresSqlConverter,
+            DuckDBSqlConverter duckDBSqlConverter)
+    {
+        this.configManager = requireNonNull(configManager, "configManager is null");
+        this.bigQuerySqlConverter = requireNonNull(bigQuerySqlConverter, "bigQuerySqlConverter is null");
```
grieve54706, line 47: Curious, why the PgFunctionBuilder is putted into Metadata, but SQLCoverter not same?

accio-main/src/main/java/io/accio/main/web/ConfigResource.java
```diff
@@ -36,11 +42,25 @@
...
+    private final Metadata metadata;
+    private final SqlConverterManager sqlConverter;
+    private final CacheServiceManager cacheService;
+    private final PgFunctionBuilderManager pgFunctionBuilderManager;
 
     @Inject
-    public ConfigResource(ConfigManager configManager)
+    public ConfigResource(
+            ConfigManager configManager,
+            Metadata metadata,
```
grieve54706, line 53: It should use `MetadataManager` for data source changed.

accio-main/src/main/java/io/accio/main/sql/SqlConverterManager.java
goldmedal: In the current architecture, PgFunctionBuilderManger doesn't need to switch the data source. All of them are handled by MetadataManager.

accio-base/src/main/java/io/accio/base/config/ConfigManager.java
```diff
@@ -291,19 +307,31 @@ public synchronized void setConfigs(List<ConfigEntry> configEntries, boolean res
                 .map(entry -> Map.entry(entry.getName(), entry.getValue()))
                 .collect(toMap(Map.Entry::getKey, entry -> entry.getValue().trim()));
         update.forEach(this::setConfigInternal);
+        boolean needReload = false;
+        for (ConfigEntry configEntry : configEntries) {
+            needReload = setConfigInternal(configEntry.getName(), configEntry.getValue());
```
grieve54706, line 312: It should use `|=`.

grieve54706 (approved):
LGTM

PR #486, opened 2024-03-14
Support query view in dynamic access mode

goldmedal:
# Description
This PR allows user to query views in dynamic access mode. The statement of view will be used to create a CTE. Every behavior of dynamic accessing only apply on the internal statement.

# Example
When dynamic accessing is enable, given view and metric
```
metric CountOrders @baseObject(Orders) {
   custkey: integer @dim
   orderstatus: varchar @dim
   totalprice: integer @measure(sum(totalprice))
}

view {
  "selectAll": "select * from CountOrders",
  "selectOneDim": "select custkey from CountOrders"
}
```

Query a SQL
```
select custkey from selectAll
```
The CountOrders will be grouped by all dimension, `custkey` and `orderstatus`.

Query another SQL
```
select custkey from selectOneDim
```
The CountOrders will be grouped by `custkey`.

### DIFF
--- a/accio-base/src/main/java/io/accio/base/sqlrewrite/AccioPlanner.java
+++ b/accio-base/src/main/java/io/accio/base/sqlrewrite/AccioPlanner.java
@@ -23,15 +23,18 @@ import io.trino.sql.tree.Statement;
 
 import java.util.List;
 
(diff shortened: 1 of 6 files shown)


grieve54706 (approved):
LGTM

PR #491, opened 2024-03-20
Add TableReference field for Model to optimize the query projection

goldmedal:
# Description
To optimize the SQL easily, we provide a new way to build a model. `TableReference` is build from `catalog`, `schema` and `table` which point to the physical table in the datasource. Instead of `refSql`, we can accurately to set which columns the SQL actually used.  This way make data source side can optimize the SQL plan more easily.

# Sample
```
{
   "name": "Orders",
   "tableReference: {
        "catalog": "memory",
        "schema": "tpch",
        "table": "orders"
    },
    columns: [
        ...
    ]
}
```

### DIFF
--- a/accio-base/src/main/java/io/accio/base/dto/TableReference.java
+++ b/accio-base/src/main/java/io/accio/base/dto/TableReference.java
@@ -17,6 +17,8 @@ package io.accio.base.dto;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
+import java.util.Objects;
+
 import static com.google.common.base.MoreObjects.toStringHelper;
 import static io.accio.base.Utils.requireNonNullEmpty;
 import static java.lang.String.format;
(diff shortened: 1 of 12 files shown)


accio-base/src/main/java/io/accio/base/dto/TableReference.java
```diff
@@ -0,0 +1,110 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+
+import static com.google.common.base.MoreObjects.toStringHelper;
+import static io.accio.base.Utils.requireNonNullEmpty;
+import static java.lang.String.format;
+
+public class TableReference
```
grieve54706, line 26: How about `CatalogSchemaTable` or using `QualifiedName` instead of it?
I thought this class was for connecting references between table and table If I saw the name only.

grieve54706 (approved):

accio-base/src/main/java/io/accio/base/dto/TableReference.java
grieve54706: This class is for the reference of the table. The name is suitable.

PR #501, opened 2024-03-21
Support wire protocol type with DuckDB

grieve54706:


wren-base/src/main/java/io/wren/base/client/duckdb/DuckdbTypes.java
```diff
@@ -109,21 +116,69 @@ public final class DuckdbTypes
...
+        }
+        catch (WrenException e) {
+            return DuckdbTypes.toPGType(metaData.getColumnTypeName(i));
+        }
+    }
+
+    public static PGType<?> toPGType(String typeName)
+    {
+        Optional<? extends PGType<?>> pgType = duckdbTypes.stream()
+                .filter(duckdbType -> duckdbType.getName().equals(getEqualTypeName(typeName)))
```
goldmedal, line 144: how about build a static hash map to lookup type? It would be efficient.

goldmedal (approved):
One minor comment. Others LGTM.

goldmedal (approved):
LGTM

PR #509, opened 2024-03-28
Implement MessagePlanner

goldmedal:
# Description
This PR implement the MessagePlanner to optimize the message flow. We categorize messages into COMMIT and PLAN. Messages of the COMMIT category will be executed immediately upon receipt. Messages of the PLAN category will not be executed immediately upon receipt but will enter the planning phase when triggered by a COMMIT. After obtaining the results of the planning, they will be executed together with the COMMIT.

There's an optimization rule for integrating `DescribePortal` and `Execute`. The planner will replace `DescribePortal` with `DescribePortalAndExecute` and `Execute` with `SendResult`. `DescribePortalAndExecute` will execute the query first and then send a `RowDescription` message according to the executing result. This approach saves one request cost to communicate with the Data Source side for each extended query. It also resolves the issue related to selecting the parameter item for DuckDB.

# Plan type Message
- Parse
- BInd
- Describe
- Execute

# Commit type Message
- Sync
- SimpleQuery
- Flush
- Close


grieve54706:
Could you list messages with `Plan` and `Commit` in the description? Or using an enum in the code base instead of a switch-case, WDYT?

goldmedal:
> Could you list messages with `Plan` and `Commit` in the description? Or using an enum in the code base instead of a switch-case, WDYT?
Sure, I'll add the list. However, instead of `enum`, I prefer use `interface` to force the developer implement the required method for them, `execute` with `Plan` and `commit` with `Commit`.
The switch-case is used to match the type char consumed from the channel. Even use `enum`, I guess we can't remove it.

wren-main/src/main/java/io/wren/main/wireprotocol/message/DescribePortalAndExecute.java
```diff
@@ -0,0 +1,102 @@
...
+                ConnectorRecordIterator connectorRecordIterable = portal.getConnectorRecordIterator();
+                FormatCodes.FormatCode[] resultFormatCodes = session.getResultFormatCodes(portalName);
+                resultSetSender = new ResultSetSender(
+                        statement,
+                        channel,
+                        connectorRecordIterable,
+                        maxRows,
+                        portal.getRowCount(),
+                        resultFormatCodes);
+                FormatCodes.FormatCode[] formatCodes = session.getResultFormatCodes(portalName);
```
grieve54706, line 93: Curious, they call the same method. Will they get different results after `new ResultSetSender`?

wren-main/src/main/java/io/wren/main/wireprotocol/message/DescribePortalAndExecute.java
```diff
@@ -0,0 +1,102 @@
...
+                ResponseMessages.sendRowDescription(channel, resultSetSender.getColumns(), formatCodes);
```
grieve54706, line 94: We change Execute to DescribePortalAndExecute. Should it not send the result set after sending the row description?

wren-main/src/main/java/io/wren/main/wireprotocol/MessagePlanner.java
```diff
@@ -0,0 +1,72 @@
...
+                Describe.DescribePortal describePortal = (Describe.DescribePortal) message;
+                portalDescriptions.put(describePortal.getTargetName(), new IndexedDescribePortal(i, describePortal));
+            }
+            if (message instanceof Execute) {
+                Execute execute = (Execute) message;
+                IndexedDescribePortal indexedDescribePortal = portalDescriptions.get(execute.getPortalName());
+                if (indexedDescribePortal != null) {
+                    DescribePortalAndExecute describePortalAndExecute = new DescribePortalAndExecute(execute);
+                    plans.set(indexedDescribePortal.index, describePortalAndExecute);
+                    plans.add(new SendResult(describePortalAndExecute));
```
grieve54706, line 50: Curious as, to why we need to split the message into two plans `DescribePortalAndExecute` and `SendResult`. Seem the `DescribePortalAndExecute` should include sending results for the naming `Execute`.

wren-main/src/main/java/io/wren/main/wireprotocol/message/SimpleQuery.java
```diff
@@ -0,0 +1,110 @@
...
+                return CompletableFuture.completedFuture(null);
+            }
+            ResultSetSender resultSetSender = new ResultSetSender(
+                    statement,
+                    channel,
+                    iterator.get(),
+                    0,
+                    0,
+                    null);
+            ResponseMessages.sendRowDescription(channel, wireProtocolSession.describePortal("").get(), null);
```
grieve54706, line 98: It should use `resultSetSender.getCloumns()` instead of `wireProtocolSession.describePortal("")`, Am I right?

wren-main/src/main/java/io/wren/main/wireprotocol/MessagePlanner.java
goldmedal: Because we cannot guarantee that "Execute" is sent immediately after "Describe". If there're something existed between them, the ordering of message should be ensured.
For example, 
```
PARSE 1,
BIND 1,
DescribePortal 1,
Parse 2, 
Execute 1
Sync
```
The response should be
```
ParseComplete
BindComplete
RowDescription
ParseComplete
DataRow
...
CommandComplete
ReadyForQuery
```

If the DescribePortalAndExecute do `SendResult`, the `DataRow` will be in front of the second ParseComplete.

wren-main/src/main/java/io/wren/main/wireprotocol/message/DescribePortalAndExecute.java
goldmedal: No, it's same. It's a duplicate calling.

wren-main/src/main/java/io/wren/main/wireprotocol/message/DescribePortalAndExecute.java
goldmedal: The reason is same as https://github.com/Canner/wren-engine/pull/509#discussion_r1544215530

wren-main/src/main/java/io/wren/main/wireprotocol/message/SimpleQuery.java
goldmedal: Yes, you're right. It can use getColumn instead.

wren-main/src/main/java/io/wren/main/wireprotocol/MessagePlanner.java
grieve54706: How about renaming it to `PendingDescribePortal` or `Waiting` without `Execute`?

goldmedal: How about `ExecuteAndSendRowDescription`? It doesn't describe portal but send description according to `Execute`.

grieve54706 (approved):

PR #510, opened 2024-03-28
Implement SQLGlot server

grieve54706:
# Description
We want to use [tobymao/sqlglot](https://github.com/tobymao/sqlglot) for converting sql to other unsupported data sources.
But its language is `Python`, so we created a minor server via [tiangolo/fastapi](https://github.com/tiangolo/fastapi).

Our test benchmark just uses TPCH to transpile and run them successfully.

We create `TestingSQLGlotServer` for the unit test to wrap up the executing process in Python.
We also temporarily wrap the `Snowflake` lib via `TestingSnowflake` for the unit test.

We put the SQLGlot server and wren-server in one container, so we must use a wrapper shell to run two processes.

## Environment variable
- SQLGLOT_PORT: default 8000
- SQLGLOT_LOG_LEVEL: default INFO
- SNOWFLAKE_URL
- SNOWFLAKE_USER
- SNOWFLAKE_PASSWORD

## SQLGlot server API
**POST** /sqlglot/transpile
Request
```json
{
  "sql": "SELECT generate_array(1, 10)",
  "read": "bigquery",
  "write": "duckdb",
}
```
Response
```json
{
  "sql": "SELECT GENERATE_SERIES(1, 10)"
}
```

## Known issues
We also found that `SQLGlot` does not support concurrent requests in Python. It will cause a `deadlock`. So we use a fixed thread pool with one thread to protect it.
```
[ERROR]   File "/tmp/testing-sqlglot-venv8819346059888102008/lib/python3.10/site-packages/sqlglot/helper.py", line 168, in apply_index_offset
[ERROR]     from sqlglot.optimizer.annotate_types import annotate_types
[ERROR]   File "/tmp/testing-sqlglot-venv8819346059888102008/lib/python3.10/site-packages/sqlglot/optimizer/annotate_types.py", line 15, in <module>
[ERROR]     from sqlglot.optimizer.scope import Scope, traverse_scope
[ERROR]   File "<frozen importlib._bootstrap>", line 1024, in _find_and_load
[ERROR]   File "<frozen importlib._bootstrap>", line 171, in __enter__
[ERROR]   File "<frozen importlib._bootstrap>", line 116, in acquire
[ERROR] _frozen_importlib._DeadlockError: deadlock detected by _ModuleLock('sqlglot.optimizer.scope') at 140680672069520
```


wren-sqlglot-converter/src/main/java/io/wren/sqlglot/converter/SQLGlotConverter.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Override
+    public String convert(String sql, SessionContext sessionContext)
+    {
+        try {
+            return sqlGlot.transpile(sql, readDialect, writeDialect);
+        }
+        catch (IOException e) {
+            throw new RuntimeException(e);
```
goldmedal, line 53:
```suggestion
            throw new WrenException(GENERIC_INTERNAL_ERROR, e);
```
Use `WrenException` instead. It's good for us to enhance the error handling in the future.

wren-sqlglot-converter/src/main/java/io/wren/sqlglot/glot/SQLGlot.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    private static void throwWebApplicationException(StringResponseHandler.StringResponse response)
+    {
+        throw new WebApplicationException(response.getBody());
+    }
+
+    public static String getPort()
+    {
+        return Optional.ofNullable(System.getenv("SQLGLOT_PORT")).orElse("8000");
```
goldmedal, line 105: Instead of `env`, I prefer to use `Config`. It would be good to trace those exist configuration.

wren-testing/src/main/java/io/wren/testing/TestingSnowflake.java
```diff
@@ -0,0 +1,56 @@
...
+    private final Session session;
+
+    public TestingSnowflake()
+    {
+        Map<String, String> properties = ImmutableMap.<String, String>builder()
+                .put("URL", getenv("SNOWFLAKE_URL"))
+                .put("USER", getenv("SNOWFLAKE_USER"))
+                .put("PASSWORD", getenv("SNOWFLAKE_PASSWORD"))
+                .put("DB", "SNOWFLAKE_SAMPLE_DATA")
+                .put("SCHEMA", "TPCH_SF1")
```
goldmedal, line 39: Is `SNOWFLAKE_SAMPLE_DATA.TPCH_SF1` the default sample data in Snowflake?

wren-testing/src/main/java/io/wren/testing/TestingSnowflake.java
```diff
@@ -0,0 +1,56 @@
...
+                .build();
+        session = Session.builder().configs(properties).create();
+    }
+
+    public void exec(String sql)
```
goldmedal, line 44: Probably, this method should return the query result.

wren-tests/src/test/java/io/wren/testing/TPCH.java
```diff
@@ -0,0 +1,232 @@
...
+import java.util.List;
+import java.util.Map;
+
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static java.lang.String.format;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.Objects.requireNonNull;
+import static org.testng.Assert.assertFalse;
+
+public class TPCH
```
goldmedal, line 34:
```suggestion
public class Tpch
```
It would be better.

pom.xml
```diff
@@ -634,6 +641,7 @@
                     <excludes>
                         <exclude>**/*jmhTest*.java</exclude>
                         <exclude>**/*jmhType*.java</exclude>
+                        <exclude>TestTPCHWithSnowflake.java</exclude>
```
goldmedal, line 644:
```suggestion
                        <exclude>TestTPCHWithSnowflake.java</exclude>
                        <exclude>TestTPCHWithDuckDB.java</exclude>

```
I think we don't need to run tpch with duckdb everytime. It's too expensive.

wren-tests/src/test/java/io/wren/testing/TPCH.java
```diff
@@ -0,0 +1,232 @@
...
+        String sql = getTpchQuery(query);
+
+        for (int i = values.length - 1; i >= 0; i--) {
+            sql = sql.replaceAll(format(":%s", i + 1), String.valueOf(values[i]));
+        }
+
+        assertFalse(sql.matches("(?s).*:[0-9].*"), "Not all bind parameters were replaced: " + sql);
+
+        return fixTpchQuery(sql);
+    }
```
goldmedal, line 206: Probably, we can use `StringSubstitutor` to refactor here. It's helpful to do replacing string.

wren-tests/src/test/java/io/wren/testing/TPCH.java
```diff
@@ -0,0 +1,232 @@
...
+                    Column.column("ps_suppkey", WrenTypes.INTEGER, null, false, "ps_suppkey"),
+                    Column.column("ps_availqty", WrenTypes.INTEGER, null, false, "ps_availqty"),
+                    Column.column("ps_supplycost", WrenTypes.DOUBLE, null, false, "ps_supplycost"),
+                    Column.column("ps_comment", WrenTypes.VARCHAR, null, false, "ps_comment"),
+                    Column.column("partkey_suppkey", WrenTypes.VARCHAR, null, true, "concat(ps_partkey, '-', ps_suppkey)")),
+            "partkey_suppkey");
+
+    public static final List<Model> MODELS = List.of(ORDERS, LINEITEM, CUSTOMER, NATION, REGION, PART, SUPPLIER, PART_SUPP);
+
+    private TPCH() {}
```
goldmedal, line 182: We do many thing in `static` scope. Consider to make TPCH be a normal class and do initialization in ctor.

wren-sqlglot-converter/src/main/java/io/wren/sqlglot/converter/SQLGlotConverter.java
```diff
@@ -0,0 +1,79 @@
...
+    public static Builder builder()
+    {
+        return new Builder();
+    }
+
+    @Override
+    public String convert(String sql, SessionContext sessionContext)
+    {
+        try {
+            return sqlGlot.transpile(sql, readDialect, writeDialect);
```
goldmedal, line 50:
```suggestion
            LOG.info("[Input sql]: %s", sql);
            return sqlGlot.transpile(sql, readDialect, writeDialect);
            LOG.info("[Dialect sql]: %s", dialectSql);
```
Just like other sql converter, we can log the before and after.

wren-testing/src/main/java/io/wren/testing/TestingSnowflake.java
grieve54706: Yes, every new account is created with the tpch schema by Snowflake.

wren-testing/src/main/java/io/wren/testing/TestingSnowflake.java
grieve54706: We don't need the results now. We can modify it in the future if we need to.

wren-tests/src/test/java/io/wren/testing/TPCH.java
grieve54706: TPCH is an abbreviation. Should keep the upper case is better.

pom.xml
grieve54706: They both should be tested in every PR. Unfortunately, we are unable to pay for Snowflake.
But the DuckDB is free and we should keep testing to ensure the sqlglot work. It only takes 11 sec.

wren-tests/src/test/java/io/wren/testing/TPCH.java
grieve54706: LOL, I use queries that do not need to replace parameters. I will remove this method.

grieve54706:
@goldmedal I found that SQLGlot does not support multiple requests simultaneously. I will add the queue to protect it.

goldmedal:
> @goldmedal I found that SQLGlot does not support multiple requests simultaneously. I will add the queue to protect it.

Isn't it thread-safe? I'm not sure it's a good idea about use queue. Can we create multiple sqlglot instance to handle the concurrent request? The queue maybe the performance bottle neck of the concurrent queries.

grieve54706:
> > @goldmedal I found that SQLGlot does not support multiple requests simultaneously. I will add the queue to protect it.
> 
> Isn't it thread-safe? I'm not sure it's a good idea about using queue. Can we create multiple sqlglot instance to handle the concurrent request? The queue maybe the performance bottle neck of the concurrent queries.

No, the problem is on the sqlglot. If we want to concurrent requests, we need to create many python servers.

goldmedal:
I tested the sqlglot with the followed code.
```python
import threading
import time
import sqlglot
sql = """
SELECT * FROM tpch.tiny.orders
"""

sql2 = """
SELECT * FROM tpch.tiny.customer
"""


def task1():
    for _ in range(10):
        print(sqlglot.transpile(sql, read="trino",write="bigquery", identify=True, pretty=True)[0])
        print("Task 1 is executing...")
        time.sleep(1)

def task2():
    for _ in range(5):
        print(sqlglot.transpile(sql2, read="trino",write="bigquery", identify=True, pretty=True)[0])
        print("Task 2 is executing...")
        time.sleep(2)

if __name__ == "__main__":
    thread1 = threading.Thread(target=task1)
    thread2 = threading.Thread(target=task2)
    
    thread1.start()
    thread2.start()
    
    thread1.join()
    thread2.join()
    
    print("All threads have finished executing.")
```

I think that it's thread-safe. Could you post the error message of the concurrent request?

grieve54706:
> I tested the sqlglot with the followed code.
> 
> I think that it's thread-safe. Could you post the error message of the concurrent request?

```
[ERROR]   File "/tmp/testing-sqlglot-venv8819346059888102008/lib/python3.10/site-packages/sqlglot/helper.py", line 168, in apply_index_offset
[ERROR]     from sqlglot.optimizer.annotate_types import annotate_types
[ERROR]   File "/tmp/testing-sqlglot-venv8819346059888102008/lib/python3.10/site-packages/sqlglot/optimizer/annotate_types.py", line 15, in <module>
[ERROR]     from sqlglot.optimizer.scope import Scope, traverse_scope
[ERROR]   File "<frozen importlib._bootstrap>", line 1024, in _find_and_load
[ERROR]   File "<frozen importlib._bootstrap>", line 171, in __enter__
[ERROR]   File "<frozen importlib._bootstrap>", line 116, in acquire
[ERROR] _frozen_importlib._DeadlockError: deadlock detected by _ModuleLock('sqlglot.optimizer.scope') at 140680672069520
```
Python is difficult to run concurrently, But Fastapi can easily reappear.

goldmedal (approved):
LGTM. Feel free to merge it after CI passed.

grieve54706:
> > I tested the sqlglot with the followed code.
> > I think that it's thread-safe. Could you post the error message of the concurrent request?
> 
> ```
> [ERROR]   File "/tmp/testing-sqlglot-venv8819346059888102008/lib/python3.10/site-packages/sqlglot/helper.py", line 168, in apply_index_offset
> [ERROR]     from sqlglot.optimizer.annotate_types import annotate_types
> [ERROR]   File "/tmp/testing-sqlglot-venv8819346059888102008/lib/python3.10/site-packages/sqlglot/optimizer/annotate_types.py", line 15, in <module>
> [ERROR]     from sqlglot.optimizer.scope import Scope, traverse_scope
> [ERROR]   File "<frozen importlib._bootstrap>", line 1024, in _find_and_load
> [ERROR]   File "<frozen importlib._bootstrap>", line 171, in __enter__
> [ERROR]   File "<frozen importlib._bootstrap>", line 116, in acquire
> [ERROR] _frozen_importlib._DeadlockError: deadlock detected by _ModuleLock('sqlglot.optimizer.scope') at 140680672069520
> ```
> 
> Python is difficult to run concurrently, But Fastapi can easily reappear.

I can't reappear this problem after making the test class with a single thread by another test case like sending 100 requests at the same time.

PR #512, opened 2024-04-02
Use import_module to pre load sqlglot to avoid concurrent load fail

grieve54706:


grieve54706:
`import_module` does not work.

PR #519, opened 2024-04-03
Upgrade to Java 21 and airbase 153

goldmedal:
# Description
- Upgrade the airbase version to 153
- Upgrade JDK to 21
- Change the base docker image from `Zulu` to `Temurin`.
- Although airlift is migrating from TestNG to JUnit, we won't follow suit in this PR. Because JUnit lacks something akin to DataProvider, which is heavily utilized in wren engine testing, it's challenging to migrate to JUnit at this time.

### DIFF
--- a/wren-main/src/main/java/io/wren/main/sqlglot/SQLGlot.java
+++ b/wren-main/src/main/java/io/wren/main/sqlglot/SQLGlot.java
@@ -35,7 +35,7 @@ import static io.airlift.http.client.JsonBodyGenerator.jsonBodyGenerator;
 import static io.airlift.http.client.Request.Builder.preparePost;
 import static io.airlift.http.client.StringResponseHandler.createStringResponseHandler;
 import static io.airlift.json.JsonCodec.jsonCodec;
-import static jakarta.ws.rs.core.MediaType.APPLICATION_JSON_TYPE;
+import static jakarta.ws.rs.core.MediaType.APPLICATION_JSON;
 import static java.util.concurrent.TimeUnit.SECONDS;
 
 public class SQLGlot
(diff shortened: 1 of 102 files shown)


grieve54706:
Do we will create an issue to make all tests with JUnit?

grieve54706:
Do they have a new code-style schema? I found some import order and `import.*` is different.

goldmedal:
> Do we will create an issue to make all tests with JUnit? JUnit5 has `@ParameterizedTest` that could be helpful.

The style of `ParameterizedTest` is very different from `DataProvider`. I'm not sure that it's a good idea to use it.

> Do they have a new code-style schema? I found some import order and `import.*` is different.

I'm not sure but I didn't import any new schema.

wren-main/src/main/java/io/wren/main/sqlglot/SQLGlot.java
```diff
@@ -86,8 +84,8 @@ public String transpile(String sql, Dialect read, Dialect write)
             throws IOException
     {
         Request request = preparePost()
-                .setUri(UriBuilder.fromUri(baseUri).path("sqlglot").path("transpile").build())
-                .setHeader(CONTENT_TYPE, APPLICATION_JSON)
+                .setUri(baseUri.resolve("sqlglot/transpile"))
+                .setHeader(CONTENT_TYPE, APPLICATION_JSON_TYPE.toString())
```
grieve54706, line 88: It could use `APPLICATION_JSON` instead of ``APPLICATION_JSON_TYPE.

grieve54706:
> The style of `ParameterizedTest` is very different from `DataProvider`. I'm not sure that it's a good idea to use it.

If we want to follow the airbase, maybe using the same test framework is a good choice.

goldmedal:
> > The style of `ParameterizedTest` is very different from `DataProvider`. I'm not sure that it's a good idea to use it.
> 
> If we want to follow the airbase, maybe using the same test framework is a good choice.

Sure, let's create an issue for it.

grieve54706 (approved):

goldmedal:
Related issue: #520

PR #524, opened 2024-04-12
Implement Snowflake data source

grieve54706:
## Description

We implement the `Snowflake` data source. We also use `sqlglot` to help us handle some statement cases.

We don't use Snowflake recommend lib `snowpark`, because it has too many features we don't need. We use `snowflake-jdbc` instead it.

The snowflake-jdbc includes `Apache-arrow` with a compatibility problem(See https://arrow.apache.org/docs/java/install.html#java-compatibility). Must add arg `--add-opens=java.base/java.nio=ALL-UNNAMED`.

We found the Snowflake SQL if you make the identifiers not delimited, Snowflake will upper-case the identifier name.

The same as before, we don't pay for Snowflake to own an account to test. So we exclude the test cases about Snowflake.

### TODO
- [ ] Test case about Type


wren-main/src/main/java/io/wren/main/connector/snowflake/SnowflakeClient.java
```diff
@@ -0,0 +1,119 @@
...
+        this.config = requireNonNull(config, "config is null");
+    }
+
+    public Connection createConnection()
+            throws SQLException
+    {
+        try {
+            Class.forName("net.snowflake.client.jdbc.SnowflakeDriver");
+            // TODO: Remove this line, it just use dependency to avoid dependency analyze failure
+            SnowflakeDriver driver = new SnowflakeDriver();
```
goldmedal, line 55: It can be fixed by chaanging the `snowflake-jdbc` dependency scope to `runtime`.

wren-main/src/main/java/io/wren/main/connector/snowflake/SnowflakeMetadata.java
```diff
@@ -0,0 +1,181 @@
...
+    {
+        throw new UnsupportedOperationException("Unnecessary method");
+    }
+
+    @Deprecated
+    @Override
+    public String getMetadataSchemaName()
+    {
+        throw new UnsupportedOperationException("Unnecessary method");
+    }
```
goldmedal, line 174: Let's remove those legacy method from `Metadata`.

wren-main/src/main/java/io/wren/main/connector/snowflake/SnowflakeRecordIterator.java
```diff
@@ -0,0 +1,162 @@
...
+
+            resultSet = statement.executeQuery();
+
+            ResultSetMetaData resultSetMetaData = resultSet.getMetaData();
+
+            this.columnCount = resultSetMetaData.getColumnCount();
+
+            this.columns = buildColumns(resultSetMetaData);
+
+            hasNext = resultSet.next();
```
goldmedal, line 72:
```suggestion
            this.connection = connection;
            statement = connection.prepareStatement(sql);
            
            for (int i = 0; i < parameters.size(); i++) {
                statement.setObject(i + 1, parameters.get(i).getValue());
            }
            
            resultSet = statement.executeQuery();
            ResultSetMetaData resultSetMetaData = resultSet.getMetaData();
            this.columnCount = resultSetMetaData.getColumnCount();
            this.columns = buildColumns(resultSetMetaData);
            hasNext = resultSet.next();
```
too many empty line.

wren-main/src/main/java/io/wren/main/connector/snowflake/SnowflakeSqlConverter.java
```diff
@@ -0,0 +1,55 @@
...
+                .build();
+    }
+
+    @Override
+    public String convert(@Language("sql") String sql, SessionContext sessionContext)
+    {
+        Node rewrittenNode = parseSql(sql);
+        LOG.info("SnowflakeSqlConverter [Input sql]: %s", sql);
+        String dialectSql = sqlGlotConverter.convert(formatSql(rewrittenNode, SNOWFLAKE), sessionContext);
+        LOG.info("SnowflakeSqlConverter [Dialect sql]: %s", dialectSql);
```
goldmedal, line 52:
```suggestion
        String dialectSql = sqlGlotConverter.convert(formatSql(rewrittenNode, SNOWFLAKE), sessionContext);
```
We have printed the log in `SqlGlotConverter`. I thinks that we don't need to print the similar one again.

wren-main/src/main/java/io/wren/main/pgcatalog/regtype/BigQueryPgMetadata.java
```diff
@@ -28,6 +28,7 @@
 import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
 
+@Deprecated
```
goldmedal, line 31: Let's remove this class

wren-tests/src/test/java/io/wren/testing/TestConfigResource.java
```diff
@@ -46,6 +46,7 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
+@Test(singleThreaded = true, description = "Every test will effect each other, so we need to run them in single thread.")
```
goldmedal, line 49: IMO, `description` is used to explain why needs this test. I prefer to explain why use `singleThreaded` by comments.
```suggestion
//  description = "Every test will effect each other, so we need to run them in single thread."
@Test(singleThreaded = true)
```

wren-tests/src/test/java/io/wren/testing/snowflake/AbstractWireProtocolTestWithSnowflake.java
```diff
@@ -0,0 +1,114 @@
...
+import static java.lang.System.getenv;
+import static java.util.Objects.requireNonNull;
+
+public abstract class AbstractWireProtocolTestWithSnowflake
+        extends AbstractWireProtocolTest
+{
+    protected TestingSQLGlotServer testingSQLGlotServer;
+
+    protected static final Manifest DEFAULT_MANIFEST = Manifest.builder()
+            .setCatalog("canner-cml")
```
goldmedal, line 48:
```suggestion
            .setCatalog("wrenai")
```

wren-tests/src/test/java/io/wren/testing/snowflake/AbstractWireProtocolTestWithSnowflake.java
```diff
@@ -0,0 +1,114 @@
...
+
+    protected Optional<String> getWrenMDLPath()
+    {
+        return Optional.of(requireNonNull(getClass().getClassLoader().getResource("snowflake/mdl.json")).getPath());
+    }
+
+    @Override
+    protected String getDefaultCatalog()
+    {
+        return "canner-cml";
```
goldmedal, line 106:
```suggestion
        return "wrenai";
```

wren-tests/src/test/java/io/wren/testing/snowflake/AbstractWireProtocolTestWithSnowflake.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Override
+    protected String getDefaultSchema()
+    {
+        return "tpch_tiny";
+    }
```
goldmedal, line 113:
```suggestion
    }
    @Override
    protected void cleanup()
    {
       // clean the generated schema
    }
```
I guess that we will create something like `pg_catalog` in Snowflake. Just like BigQuery, we should clean up it after tested.

wren-tests/src/test/resources/snowflake/mdl.json
```diff
@@ -0,0 +1,388 @@
+{
+  "catalog": "canner-cml",
```
goldmedal, line 2:
```suggestion
  "catalog": "wrenai",
```

wren-main/src/main/java/io/wren/main/connector/snowflake/SnowflakeSqlConverter.java
grieve54706: We format the sql with SNOWFLAKE dialect via `SqlFormatter`. Or we should split the two methods covert and format?

goldmedal: hmmm.. oh I found a problem. We shouldn't format SQL by `SNOWFLAKE` dialect here. We should always provide `trino` SQL to `SqlGlot` to get `Snowflake` SQL. We should use `DEFAULT` here.
```suggestion
        String dialectSql = sqlGlotConverter.convert(formatSql(rewrittenNode, DEFAULT), sessionContext);
```
Then, the formatted SQL will printed by `SqlGlotConverter`. We don't need to print it twice.

wren-tests/src/test/java/io/wren/testing/bigquery/AbstractWireProtocolTestWithBigQuery.java
```diff
@@ -88,16 +83,4 @@ protected String getDefaultSchema()
     {
         return "tpch_tiny";
     }
-
-    @Override
-    protected void cleanup()
-    {
-        try {
-            BigQueryMetadata metadata = getInstance(Key.get(BigQueryMetadata.class));
-            BigQueryClient bigQueryClient = metadata.getBigQueryClient();
```
goldmedal, line 97: There're are a bug. We will create a pg_catalog in BigQuery. We should remove it after test. 
```
bigQueryClient.dropDatasetWithAllContent(DatasetId.of(getDefaultCatalog(), metadata.getPgCatalogName()));
```

grieve54706: It was lost due to #528. I will add a commit to fix it.

goldmedal (approved):
LGTM

PR #528, opened 2024-04-17
Create view metadata in pg_catalog

goldmedal:
# Description
To get the view metadata by PG wire protocol, we should create the mock table in pg metastore.

# Changed
- Describe view statement and create table in pg_catalog
- Remove legacy metadata schema name.

# Known issue
- Get wrong type by JDBC column metadata

### DIFF
--- a/wren-main/src/main/java/io/wren/main/pgcatalog/PgCatalogManager.java
+++ b/wren-main/src/main/java/io/wren/main/pgcatalog/PgCatalogManager.java
@@ -141,20 +141,24 @@ public class PgCatalogManager
                         mdl.getColumnType(metric.getName(), metric.getWindow().getName()));
                 sb.append(format("CREATE TABLE IF NOT EXISTS \"%s\".\"%s\" (%s);\n", mdl.getSchema(), metric.getName(), cols));
             });
-            mdl.listViews().stream().map(view -> {
-                try {
-                    return Optional.of(new DescribedView(view.getName(), previewService.dryRun(mdl, view.getStatement()).join()));
-                }
-                catch (Exception e) {
-                    LOG.error(e, "Failed to describe view %s", view.getName());
-                    return Optional.empty();
-                }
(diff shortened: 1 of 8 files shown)


wren-main/src/main/java/io/wren/main/pgcatalog/PgCatalogManager.java
```diff
@@ -135,6 +141,20 @@ public void syncPgMetastore()
...
+                catch (Exception e) {
+                    LOG.error(e, "Failed to describe view %s", view.getName());
+                    return Optional.empty();
+                }
+            }).forEach(describedViewOptional -> {
+                describedViewOptional.map(view -> (DescribedView) view).ifPresent(describedView -> {
+                    String cols = describedView.columns.stream().map(column -> format("\"%s\" %s", column.getName(), column.getType().typName())).collect(joining(","));
+                    sb.append(format("CREATE TABLE IF NOT EXISTS \"%s\".\"%s\" (%s);\n", mdl.getSchema(), describedView.name, cols));
+                });
+            });
```
grieve54706, line 157: May we can split the map and ifPresent like this.
```java
mdl.listViews().stream().map(view -> {
    try {
        return Optional.of(new DescribedView(view.getName(), previewService.dryRun(mdl, view.getStatement()).join()));
    }
    catch (Exception e) {
        LOG.error(e, "Failed to describe view %s", view.getName());
        return Optional.empty();
    }
})
.filter(Optional::isPresent)
.map(Optional::get)
.map(view -> (DescribedView) view)
.forEach(describedView -> {
    String cols = describedView.columns.stream().map(column -> format("\"%s\" %s", column.getName(), column.getType().typName())).collect(joining(","));
    sb.append(format("CREATE TABLE IF NOT EXISTS \"%s\".\"%s\" (%s);\n", mdl.getSchema(), describedView.name, cols));
});
```

grieve54706 (approved):

PR #534, opened 2024-04-24
Support type for Snowflake

grieve54706:
We implement types of the Snowflake. We found the difference from snowflake-jdbc.

- ARRAY: the snowflake-jdbc only provides string type.
- BIGINT: synonymous with NUMBER
- INT: synonymous with NUMBER
- SMALLINT: synonymous with NUMBER
- DECIMAL: synonymous with NUMBER
- REAL: synonymous with FLOAT
- BYTEA: snowflake does not cast string to varbinary.
- TIMESTAMP_NTZ: If we use `getTimestamp`, the snowflake-jdbc will adjust the timestamp with the JVM timezone. But it does not consider DST(Daylight saving time). But If we use `getString`, the date-formatter will fix the DST. We found the default formatter is with fractions 3, but we support fractions 6 and we need to set the new formatter.

We also adjust the parsing option of sql parser to `AS_DOUBLE` from `AS_DECIMAL` for exact numeric.


wren-cache/src/main/java/io/wren/cache/CacheManager.java
```diff
@@ -190,7 +190,7 @@ private CompletableFuture<Void> doCache(AnalyzedMDL analyzedMDL, CacheInfo cache
                     format("select * from %s", cacheInfo.getName()),
                     sessionContext,
                     analyzedMDL);
-            Statement parsedStatement = sqlParser.createStatement(wrenRewritten, PARSE_AS_DECIMAL);
+            Statement parsedStatement = sqlParser.createStatement(wrenRewritten, PARSE_AS_DOUBLE);
```
goldmedal, line 193:
```suggestion
            Statement parsedStatement = parseSql(wrenRewritten);
```

wren-main/src/main/java/io/wren/main/wireprotocol/WireProtocolSession.java
```diff
@@ -287,7 +287,7 @@ private void parseMetastoreSemiQuery(String statementName, String statement, Lis
                 .build();
         try {
             Statement metadataQueryStatement = MetastoreSqlRewrite.rewrite(regObjectFactory,
-                    sqlParser.createStatement(statementPreRewritten, PARSE_AS_DECIMAL));
+                    sqlParser.createStatement(statementPreRewritten, PARSE_AS_DOUBLE));
```
goldmedal, line 290:
```suggestion
                    parseSql(statementPreRewritten);
```

wren-main/src/main/java/io/wren/main/wireprotocol/WireProtocolSession.java
```diff
@@ -313,7 +313,7 @@ private void parseDataSourceQuery(String statementName, String statement, List<I
                 analyzedMDL);
         // TODO: support set session property
         // validateSetSessionProperty(statementPreRewritten);
-        Statement parsedStatement = sqlParser.createStatement(wrenRewritten, PARSE_AS_DECIMAL);
+        Statement parsedStatement = sqlParser.createStatement(wrenRewritten, PARSE_AS_DOUBLE);
```
goldmedal, line 316:
```suggestion
        Statement parsedStatement = parseSql(wrenRewritten);
```

wren-testing/src/main/java/io/wren/testing/AbstractTestFramework.java
```diff
@@ -83,7 +83,7 @@ protected void cleanup() {}
 
     protected List<List<Object>> query(@Language("SQL") String sql)
     {
-        sql = formatSql(SQL_PARSER.createStatement(sql, new ParsingOptions(AS_DECIMAL)), DUCKDB);
+        sql = formatSql(SQL_PARSER.createStatement(sql, new ParsingOptions(AS_DOUBLE)), DUCKDB);
```
goldmedal, line 86:
```suggestion
        sql = formatSql(parseSql(sql), DUCKDB);
```

wren-main/src/main/java/io/wren/main/connector/snowflake/SnowflakeClient.java
```diff
@@ -63,7 +64,13 @@ public Connection createConnection()
         config.getWarehouse().ifPresent(warehouse -> props.setProperty("warehouse", warehouse));
         config.getRole().ifPresent(role -> props.setProperty("role", role));
 
-        return DriverManager.getConnection(config.getJdbcUrl(), props);
+        Connection connection = DriverManager.getConnection(config.getJdbcUrl(), props);
+
+        Statement statement = connection.createStatement();
+        statement.execute("ALTER SESSION SET TIMEZONE='UTC'");
```
goldmedal, line 70: Which is the default timezone?

wren-main/src/main/java/io/wren/main/connector/snowflake/SnowflakeSqlConverter.java
```diff
@@ -53,8 +54,9 @@ public String convert(@Language("sql") String sql, SessionContext sessionContext
 
         LOG.info("[Input sql]: %s", sql);
 
-        // TODO: Add more rewrites
-        List<SqlRewrite> sqlRewrites = ImmutableList.of();
+        // We may use custom dialect of SqlGlot instead of this
+        List<SqlRewrite> sqlRewrites = ImmutableList.of(
+                RewriteCast.INSTANCE);
```
goldmedal, line 59: Could you add some comment for the purpose of this rule?

wren-main/src/main/java/io/wren/main/sql/snowflake/RewriteCast.java
```diff
@@ -0,0 +1,109 @@
...
+    @Override
+    public Node rewrite(Node node, Metadata metadata)
+    {
+        return new CastRewriter().process(node, null);
+    }
+
+    private static class CastRewriter
+            extends BaseRewriter<Void>
+    {
+        private boolean targetDataTypeIsBytea;
```
goldmedal, line 49: We can't set the flag to be a global variable because the child of a cast could be not only `StringLiteral` but others, such as a function call. Consider a case:
```
SELECT CAST(reverse('123') AS bytea) c1
```
If we use a global variable to handle this flag, the result will be
```
SELECT CAST(reverse(HEX_ENCODE('123')) AS bytea) c1
```
However, the expected result is 
```
SELECT CAST(HEX_ENCODE(reverse('123')) AS bytea) c1
```

goldmedal (commented):
Looks great, but there are some comments. I'll review the remaining part next week.

goldmedal:
No more suggestions for now. Others LGTM.

wren-base/src/main/java/io/wren/base/sqlrewrite/WrenPlanner.java
```diff
@@ -46,10 +46,10 @@ public static String rewrite(String sql, SessionContext sessionContext, Analyzed
 
     public static String rewrite(String sql, SessionContext sessionContext, AnalyzedMDL analyzedMDL, List<WrenRule> rules)
     {
-        Statement statement = SQL_PARSER.createStatement(sql, new ParsingOptions(AS_DECIMAL));
+        Statement statement = SQL_PARSER.createStatement(sql, new ParsingOptions(AS_DOUBLE));
```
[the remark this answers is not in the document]
grieve54706: That sounds great. I love it.

wren-main/src/main/java/io/wren/main/connector/snowflake/SnowflakeClient.java
grieve54706: The default is `America/Los_Angeles`.

goldmedal (approved):
LGTM

PR #540, opened 2024-05-03
Enhance null value handling

goldmedal:
# Description
Enhance the null result for each connector. I haven't tested the Snowflake connector. Just add some handling and test case.


grieve54706:
The new test case is successful in the Snowflake, Thanks.

grieve54706 (approved):

PR #541, opened 2024-05-03
Fix the expression of metric dimension generating

### LINKED ISSUE #537: The expression of a dimension in metric can't invoke a function.
Consider a metric case
```json
    {
      "name": "RevenueYear",
      "baseObject": "orders",
      "dimension": [
        {
          "name": "PurchaseYear",
          "type": "DATE",
          "isCalculated": false,
          "notNull": false,
          "expression": "date_trunc('YEAR', PurchaseTimestamp)"
        }
      ],
      "measure": [
        {
          "name": "Price",
          "type": "DOUBLE",
          "expression": "sum(order_items.Price)"
        }
      ]
    }
```

The generated SQL is incorrect:
```
"RevenueYear" AS (
   SELECT
     orders.date_trunc('YEAR', PurchaseTimestamp) "PurchaseYear"
   , sum("orders_relationsub"."Price") "Price"
   FROM
     ((
      SELECT *
      FROM
        (
         SELECT *
         FROM
           "orders"
      )  "orders"
   )  "orders"
   LEFT JOIN (
      SELECT
        "orders"."OrderId"
      , "order_items"."Price"
      FROM
        ("orders"
      LEFT JOIN "order_items" ON ("orders"."OrderId" = "order_items"."OrderId"))
   )  "orders_relationsub" ON ("orders"."OrderId" = "orders_relationsub"."OrderId"))
   GROUP BY 1
) 
```

`date_trunc` should not be added prefix.

goldmedal:
fixed #537

### DIFF
--- a/wren-base/src/main/java/io/wren/base/sqlrewrite/MetricSqlRender.java
+++ b/wren-base/src/main/java/io/wren/base/sqlrewrite/MetricSqlRender.java
@@ -16,7 +16,6 @@ package io.wren.base.sqlrewrite;
 
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
-import io.trino.sql.SqlFormatter;
 import io.trino.sql.tree.DereferenceExpression;
 import io.trino.sql.tree.Expression;
 import io.trino.sql.tree.ExpressionRewriter;
@@ -41,6 +40,7 @@ import java.util.stream.IntStream;
 
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
+import static io.trino.sql.SqlFormatter.formatSql;
 import static io.wren.base.sqlrewrite.Utils.parseExpression;
 import static io.wren.base.sqlrewrite.Utils.parseQuery;
 import static java.lang.String.format;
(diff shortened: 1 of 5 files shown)


wren-base/src/main/java/io/wren/base/sqlrewrite/MetricSqlRender.java
```diff
@@ -151,21 +151,31 @@ protected String getSelectItemsExpression(Column column, Optional<String> relati
...
+            @Override
+            public Expression rewriteIdentifier(Identifier node, Void context, ExpressionTreeRewriter<Void> treeRewriter)
+            {
+                if (baseModel.getColumns().stream().anyMatch(c -> c.getName().equalsIgnoreCase(node.getValue()))) {
+                    return new DereferenceExpression(new Identifier(baseModel.getName(), true), new Identifier(node.getValue(), true));
+                }
+                return node;
+            }
+        }, expression);
     }
```
grieve54706, line 179: If they have same pattern to format. how about extract the model string like below:
```java
  String model;
  if (!relationshipInfos.isEmpty() && relationableBase.isPresent()) {
      model = RelationshipRewriter.relationshipAware(relationshipInfos, relationableBase.get(), expression).toString();
  }
  else if (isMeasure) {
      model = awareModel(column.getExpression().orElseThrow(() -> new IllegalArgumentException("measure column must have expression")), baseModel);
  }
  else {
      model = awareModel(column.getSqlExpression(), baseModel);
  }
  return format("%s AS \"%s\"", model, column.getName());
```

grieve54706 (approved):
LGTM, minor refactor.

wren-base/src/main/java/io/wren/base/sqlrewrite/MetricSqlRender.java
goldmedal: The return result of awareModel is a written expression. Like `orderkey` would be `Orders.orderkey`. I think that the current way is better.

PR #544, opened 2024-05-07
Fix the DuckDB type mapping and enhance the error handling for MDL deploying

goldmedal:
# Description
- add missing type mapping for PGType to DuckDB Type
- Handle the error for known pg type which don't supported by DuckDB
- Return http 400 error when mdl deploying failed.


grieve54706 (approved):
LGTM, Thanks.

PR #558, opened 2024-05-21
Adjust with fastapi cli

grieve54706:
Because fastapi release [0.111.0](https://github.com/tiangolo/fastapi/releases/tag/0.111.0) include the [FastAPI CLI](https://fastapi.tiangolo.com/fastapi-cli/). We can use the CLI to start the fastapi service.


goldmedal (approved):
LGTM

PR #559, opened 2024-05-21
Add port setting

grieve54706:


goldmedal:
Previously, we can just use
```
IBIS_PORT=8080 make start
```
To change the binding port for fastapi server.

Is there any concern for this way?

grieve54706:
> Previously, we can just use
> 
> ```
> IBIS_PORT=8080 make start
> ```
> 
> To change the binding port for fastapi server.
> 
> Is there any concern for this way?

Before, we start service via `uvicorn.run` in the `main.py` with custom env var. But #558 adjusted the starter via `fastapi-cli` that provides `---port` to change the port. We change the env var name to `PORT`. And you want to put it before the command is also worked.

goldmedal:
ok, I didn't notice you create a feature branch `feature/ibis-server`. Just compare the difference between `main` and it. Now I got it.
I think the purpose of this PR is add an env to config the binding port of fastapi. Could you modify the PR description?

goldmedal (approved):
Others LGTM.

PR #562, opened 2024-05-22
Adjust data query api to receive Wren SQL

grieve54706:
Make the API receive Wren SQL then rewrite the SQL to data source sql via `dry-plan` API of Wren Engine.


ibis-server/app/model/mdl.py
```diff
@@ -0,0 +1,82 @@
+from enum import StrEnum, auto
+from typing import List, Optional, Dict
+
+from pydantic import BaseModel, Field
+
+
+class Manifest(BaseModel):
```
goldmedal, line 7: I think we should change the dry-plan API in Wren Engine.
```json
{
   "manifest_json": "....",   // just json string
   "sql": "...",
   "modelingOnly": true
}
```

We don't need handle the DTOs everywhere.

ibis-server/app/config.py
```diff
@@ -0,0 +1,16 @@
+import os
+
+from dotenv import load_dotenv
+
+
+class Config:
+    def __init__(self):
+        load_dotenv(override=True)
+        self.wren_engine_endpoint = os.getenv('WREN_ENGINE_ENDPOINT')
```
goldmedal, line 9: Seems the env is required. If have a check when server starting, it's would be better.

ibis-server/app/model/data_source.py
```diff
@@ -24,31 +26,60 @@ def get_connection(self, dto):
...
+            user=info.user,
+            password=info.password,
+            account=info.account,
+            database=info.database,
+            schema=info.sf_schema,
         )
+
+
+class PostgresConnectionInfo(BaseModel):
+    host: str = Field(examples=["localhost"])
```
goldmedal, line 60: I think we should also support to use [the connection string of postgres](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING-KEYWORD-VALUE). I found ibis-postgres has [another connection](https://ibis-project.org/backends/postgresql#ibis.connect-url-format) mode.

Because the user may have additional config in the connection(e.g. ssl-enable), it's not enough if we only provide host, port, database, user and password.

ibis-server/app/model/data_source.py
```diff
@@ -24,31 +26,59 @@ def get_connection(self, dto):
...
         )
+
+
+class PostgresConnectionInfo(BaseModel):
+    host: Optional[str] = Field(examples=["localhost"], default=None)
+    port: Optional[int] = Field(default=5432)
+    database: Optional[str] = None
+    user: Optional[str] = None
+    password: Optional[str] = None
+    jdbc_url: Optional[str] = None
```
goldmedal, line 63:
```suggestion
    connection_string: Optional[str] = None
```
It's not a jdbc url actually.

ibis-server/app/model/data_source.py
```diff
@@ -24,31 +26,59 @@ def get_connection(self, dto):
...
-        return ibis.postgres.connect(
-            user=dto.user,
-            password=dto.password,
-            database=dto.database,
-            host=dto.host,
-            port=dto.port,
-        )
+    def get_postgres_connection(info: 'PostgresConnectionInfo') -> BaseBackend:
+        if info.jdbc_url:
+            resource = info.jdbc_url.removeprefix("jdbc:")
```
goldmedal, line 31: I think user should not provide jdbc prefix here.

goldmedal (approved):
Thanks @grieve54706. LGTM.

PR #572, opened 2024-05-28
Fix v1-engine-changed label event

grieve54706:
The workflow `build.yml` was not triggered at PR updated.

The workflow `integration-tests.yml` was triggered no matter what label name because the syntax was wrong.


grieve54706:
I tried the `on.pull_request.types: synchronize, edited` does not work. Pending this issue #578 in the future.

PR #575, opened 2024-05-28
Add logger

grieve54706:
I added a `logger.py` to wrap the `logging` for basic config like log level could be one side.

Before we import logging then use `getLogger(name)`, we should from `app.logger` import `get_logger` instead of it.


goldmedal:
Could you add some example to show how to use this logger?

goldmedal (approved):
LGTM

PR #581, opened 2024-05-29
Adjust manifest string to base64

grieve54706:
Before the manifest string was a JSON string, but it could have some issues about escaping character.
Making the string base64 may be better.


goldmedal (approved):
LGTM. Thanks @grieve54706

PR #582, opened 2024-05-29
Add heap_size setting env for engine image

goldmedal:
# Description
Add two env to set the heap size of JVM.
- MAX_HEAP_SIZE
- INI_HEAP_SIZE

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -1,6 +1,9 @@
+#!/bin/bash
+export ENV_MAX_HEAP_SIZE=$2
+export ENV_INI_HEAP_SIZE=$3
 python3 wren-sqlglot-server/main.py &
 
 # Required add-opens=java.nio=ALL-UNNAMED for Apache arrow in the Snowflake
-java -Dconfig=etc/config.properties \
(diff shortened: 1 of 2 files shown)


docker/entrypoint.sh
```diff
@@ -1,6 +1,9 @@
+#!/bin/bash
+export ENV_MAX_HEAP_SIZE=$2
+export ENV_INI_HEAP_SIZE=$3
 python3 wren-sqlglot-server/main.py &
 
 # Required add-opens=java.nio=ALL-UNNAMED for Apache arrow in the Snowflake
-java -Dconfig=etc/config.properties \
+java -Xmx${ENV_MAX_HEAP_SIZE:-"512m"} -Xms${ENV_INI_HEAP_SIZE:-"64m"}  -Dconfig=etc/config.properties \
```
grieve54706, line 7: Curious, why use `INI` not `MIN`?

goldmedal: Initial heap size.

goldmedal: hmm.. maybe change to `MIN` would be more clear.

grieve54706 (approved):
LGTM

PR #585, opened 2024-05-29
Remove unused test-with-secrets workflow

goldmedal:
# Description
This workflow isn't finished but always triggered when comments. Let's remove it and add it back when we need it.


grieve54706 (approved):
LGTM

PR #587, opened 2024-05-31
Enable pg-wire-protocol in example

goldmedal:
# Description
Due to #546 , pg-wire is disable in default. However, `psql` is a great interface for user to interact with the engine directly. Let's enable it in the example.


grieve54706 (approved):
LGTM, If they could be with an empty line in the end will be better.

goldmedal:
Thanks @grieve54706

PR #592, opened 2024-06-03
Use exec-jar instead of docker image

grieve54706:
Compared to using docker pull to get an image, using the build jar can save a lot of time and may also have requirements for modifying V2 API in the Java engine.

Using docker pull takes 9 mins, build jar only 1 min.


grieve54706:
We have observed that sometimes pulling a Docker image is very fast. We should continue to monitor this CI process for another week.

PR #593, opened 2024-06-04
Add dryRun parameter for query API

grieve54706:
Add `dryRun` parameter for query API and split `ibis.py` by data source.

Example:
```
POST /v2/ibis/postgres/query?dryRun=true
```
If successful, it will return HTTP status code `204`. If it fails, it will return HTTP status code `422`.


ibis-server/app/model/coordinator.py
```diff
@@ -0,0 +1,33 @@
+from json import loads
+
+from app.mdl.rewriter import rewrite
+from app.model.data_source import DataSource, ConnectionInfo
+
+
+class Coordinator:
```
goldmedal, line 7:
```suggestion
class Connector:
```
How about calling it `Connector`? It's not a distributed architecture. `Coordinator` might not be suitable.

goldmedal (approved):
Thanks @grieve54706. Only one naming issue others LGTM.

ibis-server/app/model/coordinator.py
grieve54706: Sure, nice name. Thanks.

PR #594, opened 2024-06-05
Add column_dtypes to forcibly convert the type

grieve54706:
## Description
We will convert types in the data frame by input dtypes.
We also format datetime to a formatted string.

## HTTP example
```
POST /v2/ibis/postgres/query
{
  "connectionInfo": {...},
  "manifestStr": "base64 string of manifest",
  "sql": 'SELECT * FROM "Orders" LIMIT 1',
  "columnDtypes": {
      "totalprice": "float",
      "orderdate": "datetime64",
      "timestamp": "datetime64",
      "timestamptz": "datetime64"
  }
}
```
## Response difference
```diff
{
    "columns": [
        "orderkey",
        "custkey",
        "orderstatus",
        "totalprice",
        "orderdate",
        "order_cust_key",
        "timestamp",
        "timestamptz"
    ],
    "data": [
        [
            1,
            370,
            "O",
            "172799.49",
-           820540800000,
+           "1996-01-02 00:00:00.000000",
            "1_370",
-           1704153599000,
+           "2024-01-01 23:59:59.000000",
-           1704153599000
+           "2024-01-01 23:59:59.000000 UTC"
        ]
    ],
    "dtypes": {
        "orderkey": "int32",
        "custkey": "int32",
        "orderstatus": "object",
-       "totalprice": "object",
+       "totalprice": "float64",
        "orderdate": "object",
        "order_cust_key": "object",
-       "timestamp": "datetime64[ns]",
+       "timestamp": "object",
-       "timestamptz": "datetime64[ns, UTC]"
+       "timestamptz": "object"
    }
}
```

## Additional information
Pandas dtypes https://pandas.pydata.org/docs/user_guide/basics.html#dtypes

### DIFF
--- a/ibis-server/tests/routers/ibis/test_bigquery.py
+++ b/ibis-server/tests/routers/ibis/test_bigquery.py
@@ -1,3 +1,6 @@
+import base64
+
+import orjson
 import pytest
 from fastapi.testclient import TestClient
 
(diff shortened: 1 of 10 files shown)


ibis-server/tests/routers/ibis/test_bigquery.py
```diff
@@ -50,22 +53,63 @@ def get_connection_info():
...
+        assert result['data'][0] == [1, 370, 'O', 172799.49, '1996-01-02 00:00:00.000000', '1_370', '2024-01-01 23:59:59.000000', '2024-01-01 23:59:59.000000 UTC']
+        assert result['dtypes'] == {
+            'orderkey': 'int64',
+            'custkey': 'int64',
+            'orderstatus': 'object',
+            'totalprice': 'float64',
+            'orderdate': 'object',
+            'order_cust_key': 'object',
+            'timestamp': 'object',
+            'timestamptz': 'object'
```
goldmedal, line 111: Curiously, why are those object after setting the specific column type?

goldmedal (commented):
Thanks @grieve54706. I have one question for this.

ibis-server/tests/routers/ibis/test_bigquery.py
grieve54706: Because we format the datetime to string. String in the dtype is objcet. The data type is mapping the data value. If we want to keep the type is datetime, we should not format the value. But we need the value format.

goldmedal: I see. Thanks. It's good for now.

goldmedal (approved):
LGTM

PR #597, opened 2024-06-06
Implement Manifest builder for testing and develop

goldmedal:
# Description
Implement the Manifest builder to improve testing. We don't need to build the test case from the file every time.

### DIFF
--- a/wren-modeling-rs/src/mdl/builder.rs
+++ b/wren-modeling-rs/src/mdl/builder.rs
@@ -1,0 +1,487 @@
+#![allow(dead_code)]
+
+use crate::mdl::manifest::{
+    Column, JoinType, Manifest, Metric, Model, Relationship, TimeGrain, TimeUnit, View,
+};
+use std::sync::Arc;
+
+/// A builder for creating a Manifest
+pub struct ManifestBuilder {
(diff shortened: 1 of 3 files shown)


goldmedal:
The integration test CI failed doesn't related to the rust side. Let's ignore it.

wren-modeling-rs/src/mdl/builder.rs
```diff
@@ -0,0 +1,487 @@
...
+                schema: "default".to_string(),
+                models: vec![],
+                relationships: vec![],
+                metrics: vec![],
+                views: vec![],
+            },
+        }
+    }
+
+    pub fn with_catalog(mut self, catalog: &str) -> Self {
```
grieve54706, line 27: How about just `catalog`?

grieve54706 (approved):

wren-modeling-rs/src/mdl/builder.rs
goldmedal: Just `catalog` may be confused with `get_catalog`. I think having a `with` prefix is better.

grieve54706 (approved):
LGTM

PR #605, opened 2024-06-07
Add validation API

grieve54706:
## Summary
- Add validation API
- Integrate DTO into the actioner.

## New API
```
POST /v2/ibis/{data_source}/validate/{rule_name}
```

### Data source
- postgres
- bigquery
- snowflake

### Rules
- column_is_valid

### Request body
```json
{
    "connectionInfo": {......},
    "manifestStr": "eyJjYXRhbG9nIjoibXlfY2F0YWxvZyIsInNjaGVtYSI6Im15X3...",
    "parameters": {
        "modelName": "Orders", 
        "columnName": "orderkey"
    }
}
```
- parameters: key-value, required. For different rules, there will be different parameters.

### Response
If the request is valid, The API will return HTTP status code `204` else return code `422`


ibis-server/app/model/querier.py
```diff
@@ -1,12 +1,19 @@
...
+from app.model.data_source import (
+    PostgresConnectionUrl,
+    PostgresConnectionInfo,
+    BigQueryConnectionInfo,
+    SnowflakeConnectionInfo,
+)
 
 
-class Connector:
+class Querier:
```
goldmedal, line 16: I think it's not a common name, and it's actually hard to pronounce. If you don't prefer `Connector`, how about `QueryAgent`?

goldmedal (approved):
Thanks, @grieve54706, nice work. Overall, LGTM. Just one naming suggestion.
BTW, I think the response format is different from the Java version #536. @onlyjackfrost may need to take care of it.

goldmedal (approved):
LGTM Thanks @grieve54706

PR #606, opened 2024-06-10
Fix the table qualified name and executable plan for DataFusion

goldmedal:
# Description
- In this PR, instead of bare table reference,  we uses full table reference to analyze MDL.
- If physical table resource is exists, we should `DefaultTableSource` instead of `LogicalTableSource`. Refer to [TableSource](https://datafusion.apache.org/library-user-guide/building-logical-plans.html#table-sources)

# Known issue
- Column expression can't work well.

### DIFF
--- a/wren-modeling-rs/core/src/logical_plan/rule.rs
+++ b/wren-modeling-rs/core/src/logical_plan/rule.rs
@@ -1,17 +1,17 @@
+use std::{collections::HashMap, fmt, fmt::Debug, sync::Arc};
 use std::cell::RefCell;
 use std::cmp::{Ordering, PartialEq};
 use std::collections::{BTreeSet, HashSet};
-use std::{collections::HashMap, fmt, fmt::Debug, sync::Arc};
 
 use arrow_schema::Field;
+use datafusion::common::{DFSchema, DFSchemaRef, Result};
 use datafusion::common::config::ConfigOptions;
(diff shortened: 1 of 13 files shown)


wren-modeling-rs/core/src/logical_plan/rule.rs
```diff
@@ -170,6 +149,40 @@ impl ModelAnalyzeRule {
...
 
+fn belong_to_mdl(mdl: &WrenMDL, table_reference: TableReference) -> bool {
+    let catalog_mismatch = table_reference
+        .catalog()
+        .map_or(true, |c| c != mdl.catalog());
+
+    let schema_mismatch = table_reference.schema().map_or(true, |s| s != mdl.schema());
+
+    catalog_mismatch || schema_mismatch
+}
```
grieve54706, line 160: IMO, using positive conditions will be better than negative. How about using `catalog_is_match` && `schema_is_match`?

grieve54706 (approved):
LGTM

wren-modeling-rs/core/src/logical_plan/rule.rs
goldmedal: hmm I agree with you. However, after I tried to change to positive condition, it triggers another qualified name bugs that it's related what I'm fixing. I think I can fix them together in next PR. Let's merge this first.

PR #609, opened 2024-06-12
Make CI to build executable jar

grieve54706:
We found sometimes we will modify the Java codebase at the same time as modifying ibis.
So we should build the jar to make sure the Java engine is the latest.

### DIFF
--- a/.github/workflows/ibis-ci.yml
+++ b/.github/workflows/ibis-ci.yml
@@ -34,6 +34,7 @@ jobs:
           mkdir etc
           echo "node.environment=production" >> etc/config.properties
           echo "wren.directory=./etc/mdl" >> etc/config.properties
+          echo "wren.experimental-enable-dynamic-fields=true" >> etc/config.properties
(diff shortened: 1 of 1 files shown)


.github/workflows/ibis-ci.yml
```diff
@@ -22,11 +22,22 @@ jobs:
...
+          java-version: '21'
+          cache: 'maven'
       - name: Start Wren JAVA engine
-        working-directory: ./example/duckdb-tpch-example
+        working-directory: .
         run: |
-          docker compose --env-file .env up -d
+          mkdir etc
+          echo "node.environment=production" >> etc/config.properties
+          echo "wren.directory=./etc/mdl" >> etc/config.properties
```
goldmedal, line 36:
```suggestion
          echo "wren.directory=./etc/mdl" >> etc/config.properties
          echo "wren.experimental-enable-dynamic-fields=true
" >> etc/config.properties
```
Let's enable dynamic query in default

goldmedal (approved):
Overalls LGTM. Thanks @grieve54706

PR #613, opened 2024-06-14
Fix the to_one relationship and enhance the testing

goldmedal:
# Changed
- The to_one relationship is available now.
- Ensure the generated plan is executable for DataFusion.
   - If table providers are registered, generate the DefaultTableSource for remote table sources.
   - If no, generate the LogicalTableSource for remote table sources. (It can't be executed by DataFusion)
- Enhance the end to end testing


wren-modeling-rs/core/tests/data/mdl.json
```diff
@@ -4,7 +4,7 @@
   "models": [
     {
       "name": "customer",
-      "tableReference": "main.customer",
+      "tableReference": "customer",
```
grieve54706, line 7: Why remove the schema?

grieve54706 (approved):

wren-modeling-rs/core/tests/data/mdl.json
goldmedal: The testing context, `RemoteContextProvider`, only register the table name without the schema. In this case, the table reference shouldn't have schema prefix.

PR #616, opened 2024-06-18
Remove rust in Dockerfile

grieve54706:
We don't use the rust in the normal flow. We can remove the failed builder for rust in the Dockerfile now.


goldmedal (approved):
Thanks @grieve54706 -- LGTM. I think the wren core python is a work in progress. We can handle this issue when we ready to integrate with ibis.

PR #619, opened 2024-06-19
Separate ibis server by programming language

grieve54706:
For a pure Python project, we extract Rust from the ibis server project. Create a new module `wren-modeling-py` for building Python lib from Rust.

We need to build a wheel by `maturin build` in `wren-modeling-py`, and then we go back `ibis-server` to install the wheel.

We provide the `just` commander `install-core` to trigger building in `wren-modeling-py` and install the wheel.
We can also use `just update-core` to apply the new change in `wren-modeling-py` or `wren-modeling-rs`.


wren-modeling-py/Cargo.toml
```diff
@@ -5,12 +5,12 @@ edition = "2021"
 include = ["/src", "pyproject.toml", "Cargo.toml", "Cargo.lock"]
 
 [lib]
-name = "modeling_core"
+name = "wren_core"
```
goldmedal, line 8:
```suggestion
name = "wren_core_py"
```
I think it's a Python adaptor. Maybe have a `py` suffix si better.

wren-modeling-py/README.md
```diff
@@ -0,0 +1,19 @@
+# Wren Modeling Core
```
goldmedal, line 1:
```suggestion
# Wren Modeling Core in Python
```
How about this? It's more clear.

ibis-server/README.md
```diff
@@ -98,3 +94,6 @@ just run
 ### Docker
 - Build the image: `just docker-build`
 - Run the container: `just docker-run`
+
+### Modeling Core
+The modeling core is written in Rust and convert to Python package. You can find the [wren-modeling-py](../wren-modeling-py) and the source code in the [wren-modeling-rs](../wren-modeling-rs) directory.
```
goldmedal, line 99:
```suggestion
The modeling core is written in Rust and provides the Python adaptor. You can find the [wren-modeling-py](../wren-modeling-py) and the source code in the [wren-modeling-rs](../wren-modeling-rs) directory.
```
How about this?

goldmedal (approved):
Thanks @grieve54706  nice works. LGMT but some minor comments.

wren-modeling-py/Cargo.toml
grieve54706: This name will be used in the Python codebase. When importing, it is weird if it is with the suffix py in the Python codebase.

goldmedal: I found datafusion-python also name it with the suffix but import without suffix.
https://github.com/apache/datafusion-python/blob/45a684445e25032961a7bb44ced3ce06f5ed9e6d/Cargo.toml#L64
Do you have any idea for it?

grieve54706: hmmm, Let me try again.

grieve54706: Cool, it seems only to follow `[tool.maturin.module-name]` in `pyproject.toml`. I will change `[lib.name]` to `wren_modeling_py` to be the same as the package name. And keep the module name `wren_core`.

PR #622, opened 2024-06-20
Separate modeling py CI

grieve54706:
One is light CI for PR, and the other is for multi-platform.

The multi-platform CI will not be used now. It is for stable release in the future.

We also make `maturin` included in poetry.


goldmedal (approved):
Thanks @grieve54706 -- Overall LGTM.

PR #623, opened 2024-06-20
Add pyo3-build-config for macos building

grieve54706:
Cargo build will be failed in the MacOS before.
This project needs to add `pyo3-build-config`.

Reference
https://pyo3.rs/v0.14.2/building_and_distribution.html#macos


goldmedal (approved):
Cool. I can build it in my local environment now. Thanks @grieve54706 LGTM.

PR #624, opened 2024-06-20
Integrate API of ibis

grieve54706:
Make data source APIs be one API. Because they are not much different.

Add `isort` into the linter list for sorting the import order.


goldmedal (approved):
Thanks @grieve54706 -- Overall LGTM.

PR #628, opened 2024-06-25
Fix the complex calculated field and rename the default schema

goldmedal:
# Changed
- Rename the default schema from `default` to `public` because the `default` is a SQL reserved word.
- Fix the calculation expression for the relationship expression and normal expression (e.g. `a.b.c1 + a1`)
- Enhance the testing for relationship


grieve54706 (approved):
LGTM

PR #625, opened 2024-06-21
Add data source MS SQL Server

grieve54706:
# Description
Make `POST /v2/ibis/{data_source}/query` support new data source `MS SQL Server`

## Request body
- host: string, `required`
- port: integer, `required`
- database: string, `required`
- user: string, `required`
- password: string, `required`
- driver: string,

> [!WARNING]
> Ibis does not support mssql using connection url. Follow https://github.com/ibis-project/ibis/issues/9415

```json
{
  "sql": "select * from \"Orders\" limit 1",
  "manifestStr": "eyJjYXRhbG9nIjoibXlfY2F0YWxvZyIsInNjaGVtYSI6Im15X3...",
  "connectionInfo": {
    "host": "localhost",
    "port": 1433,
    "database": "dbname",
    "user": "username",
    "password": "password"
  }
}
```

## Response body
- columns
- data
- dtypes: Type through the pandas

```json
{
    "columns": [
        "o_orderkey",
        "o_custkey",
        "o_orderstatus",
        "o_totalprice",
        "o_orderdate",
        "o_orderpriority",
        "o_clerk",
        "o_shippriority",
        "o_comment"
    ],
    "data": [
        [
            1,
            370,
            "O",
            172799.49,
            820540800000,
            "5-LOW",
            "Clerk#000000951",
            0,
            "nstructions sleep furiously among "
        ]
    ],
    "dtypes": {
        "o_orderkey": "int32",
        "o_custkey": "int32",
        "o_orderstatus": "object",
        "o_totalprice": "object",
        "o_orderdate": "object",
        "o_orderpriority": "object",
        "o_clerk": "object",
        "o_shippriority": "int32",
        "o_comment": "object"
    }
}
```

PR #629, opened 2024-06-25
Add FreeTDS driver in Dockerfile

grieve54706:
Follow #625, we miss the odbc driver in the docker.

### DIFF
--- a/ibis-server/Dockerfile
+++ b/ibis-server/Dockerfile
@@ -27,9 +27,32 @@ RUN poetry install --without dev
 FROM python:3.11-slim-buster as runtime
 
 # libpq-dev is required for psycopg2
+# unixodbc-dev and freetds-dev are required for pyodbc
 RUN apt-get update  \
     && apt-get -y install libpq-dev \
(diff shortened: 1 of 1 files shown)


ibis-server/Dockerfile
```diff
@@ -27,10 +27,33 @@ RUN poetry install --without dev
 FROM python:3.11-slim-buster as runtime
 
 # libpq-dev is required for psycopg2
+# unixodbc-dev and freetds-dev are required for pyodbc
 RUN apt-get update  \
     && apt-get -y install libpq-dev \
+    && apt-get -y install unixodbc-dev freetds-dev tdsodbc \
```
goldmedal, line 33: How about having a comment for `tdsodbc`? Maybe say it's required for MSSQL?

goldmedal (approved):
Thanks @grieve54706. I only have one minor comment. Others LGTM.

ibis-server/Dockerfile
grieve54706: Yes, freetds need tdsodbc. I will add tdsodbc to the comment.

PR #633, opened 2024-06-26
Modify README and justfile

grieve54706:


goldmedal (approved):
Thanks @grieve54706 LGTM.

Issue #584, opened 2024-05-29
[EPIC] Ibis-project as the data source connector


goldmedal:
# Purpose
Previously, Wren Engine implemented the data source connector by ourselves. To support a new data source, we had to spend a lot of time solving various issues, such as type mapping, syntax support, or data serialization/deserialization. [Ibis](https://github.com/ibis-project/ibis) is a Python DataFrame library that can connect to and convert SQL for many data sources.

We plan to change Wren Engine to a Python project. The core features will be implemented in Rust, and the data flow will be controlled by a Python server.

# Architecture
The architecture of the new Wren Engine is illustrated below, and it also demonstrates how to integrate with WrenAI.  
![wren-engine-ibis-server](https://github.com/Canner/wren-engine/assets/6974913/035bad65-5f77-4743-a637-5b34237dbac2)

In the meantime, I'm working on redesigning the model module in Rust and DataFusion. PR #570 is the first step in this effort. I will file another epic issue for this work later.

Because the work of redesigning the modeling module will take some time, the first version of the Ibis connector will transform WrenSQL using the original SQL plan API (referred to as `dry-plan` in [MDLResourceV2](https://github.com/Canner/wren-engine/blob/main/wren-main/src/main/java/io/wren/main/web/MDLResourceV2.java)).

# Work Lists
There're some subtasks in this epic. We should file other tickets for them.
- [x] Data Query API: Query the data source the result. @grieve54706 is working on it now.
  - #574 
  - https://github.com/Canner/WrenAI/pull/306
  - https://github.com/Canner/wren-engine/pull/594
- [x] Dry Run API: Check if the query is valid and got the schema of the result set. Same as the original `dry-run` api in [MDLResource](https://github.com/Canner/wren-engine/blob/main/wren-main/src/main/java/io/wren/main/web/MDLResource.java).
  - #593  
- [x] Analysis API: analyze a SQL to get some important information for AI service. Invoke the original analysis API in [AnalysisResource](https://github.com/Canner/wren-engine/blob/main/wren-main/src/main/java/io/wren/main/web/AnalysisResource.java) 
- [x] MDL Validation API: Provide a way to check if a MDL is valid. Same as the validation API in [MDLResource](https://github.com/Canner/wren-engine/blob/main/wren-main/src/main/java/io/wren/main/web/MDLResource.java).
  - #605
- [x] CI workflow and code quality check
  - #583

goldmedal:
This epic is totally finished.

PR #639, opened 2024-06-28
Remove the legacy connector, Cache server and PG Wire Protocol 

goldmedal:
# Description
After #584, we move all connector but DuckDB to the Ibis version. Cache and PG Wire protocol are also deprecated. In this PR, we formally remove those legacy code. The java side of Wren Engine only remain the follow features:
- WrenMDL Planner.
- Dry Plan API
- DuckDB Connector (with the connection pool)
- Preview, Dry-Run, and Validation API (for DuckDB)
- SQL analysis API for AI Decision Point
- Wren Engine no longer keep the Deployed MDL file. It means you don't need to deploy any file here but you should always carry the MDL when calling any API.


grieve54706 (approved):
You’re a lifesaver. Love you.

PR #644, opened 2024-07-01
Add query parameter limit

grieve54706:
Add new query parameter `limit` for query API to limit data size.

POST /v2/ibis/{data_source}/query?`limit=1`


ibis-server/app/routers/v2/ibis.py
```diff
@@ -27,12 +27,15 @@ def query(
     data_source: DataSource,
     dto: QueryDTO,
     dry_run: Annotated[bool, Query(alias="dryRun")] = False,
+    limit: int = 100,
```
goldmedal, line 30: I think the default value could be `unlimit`. It can also be used by the user who want to get all result.

goldmedal (approved):
Thanks @grieve54706. LGTM.

PR #645, opened 2024-07-01
Introduce the View of MDL

goldmedal:
# Description

## Introduce View
Now, we can support the view in the MDL.
```
  "views": [
    {
      "name": "customer_view",
      "statement": "select * from test.test.customer"
    }
```
User can store the Wren SQL as a view statement.
### There're some limitations:
- The table name should be qualified ( full catalog.schema.table name )
- A complex expression should have alias name. There's an issue for the complex expression. The SQL as below can't be supported currently.
```sql
select order_id, sum(price) from wrenai.public.order_items group by order_id
```

## Include SessionContext
In this PR, I include the `Datafusion::SessionContext` in the SQL planning flow. It will be responsible for store what we can use when planning SQL. I create another API named `transform_sql_with_ctx` that accepts a `SessionContext` which might be registered some tables.  The original API `transform_sql` will create a default SessionContext for planning.


grieve54706 (approved):
LGTM

PR #648, opened 2024-07-03
implement clickhouse metadata APIs

onlyjackfrost:
### Request payload Example
```json
{
  "connectionInfo": {
      "connectionUrl": "clickhouse://user:pwd@host:port/database?secure=1"
  }
}
```
 or
``` json
{
  "connectionInfo": {
      "host": "localhost",
      "port":  8443,
      "user": "my-user",
      "password": "my-pwd",
      "database": "my-database"
  }
}
```
### Response Example
- name: unique table name (might contain schema name, depends on are you listing table across schema or not )
- columns
    - name: column name in the datasource
    - type: column data type
    - notNull: boolean, nullable or not
    - description: column description(comment) if any
    - properties: column properties if any
- description: table description if any
- properties
    - schema: schema name to build tableReference
    - catalog: catalog name to build tableReference
    - table: table name to build tableReference
- primaryKey: the column name which is bind with primary constraint
```json
[
    {
        "name": "dbo.REGION",
        "columns": [
            {
                "name": "R_COMMENT",
                "type": "VARCHAR",
                "notNull": false,
                "description": "",
                "properties": null
            }
        ],
        "description": "",
        "properties": {
            "schema": "db",
            "catalog": "def",
            "table": "REGION"
        },
        "primaryKey": ""
    },
]
```

### Request payload Example
```json
{
  "connectionInfo": {
      "connectionUrl": "clickhouse://user:pwd@host:port/database"
  }
}
```
Note: ClickHouse does not support foreign key constraints.  This API will return an empty array for now.
### Response Example
```
[]
```


grieve54706:
How about adding a test for getting constraints that will return an empty array?

grieve54706 (approved):
LGTM

PR #655, opened 2024-07-04
Add dry-plan

grieve54706:
Add two APIs for the dry plan

One API only rewrites SQL with manifest
```
POST /v2/ibis/dry-plan
{
  "manifestStr": "eyJjYXRhbG9nIjoibXlfY2F0YWxvZyIsInN......",
  "sql": "SELECT orderkey, order_cust_key FROM \"Orders\" LIMIT 1"
}
```
```
WITH
    "Orders" AS (
        SELECT
            "Orders"."orderkey" AS "orderkey",
            "Orders"."custkey" AS "custkey",
            "Orders"."order_cust_key" AS "order_cust_key"
        FROM
            (
                SELECT
                    "Orders"."orderkey" AS "orderkey",
                    "Orders"."custkey" AS "custkey",
                    "Orders"."order_cust_key" AS "order_cust_key"
                FROM
                    (
                        SELECT
                            o_orderkey AS "orderkey",
                            o_custkey AS "custkey",
                            CONCAT (o_orderkey, '_', o_custkey) AS "order_cust_key"
                        FROM
                            (
                                SELECT
                                    o_orderkey,
                                    o_custkey
                                FROM
                                    public.orders
                            ) AS "Orders"
                    ) AS "Orders"
            ) AS "Orders"
    )
SELECT
    orderkey,
    order_cust_key
FROM
    "Orders"
LIMIT
    1
```

Another API will rewrite SQL with manifest and transpile by data source dialect.
```
POST /v2/ibis/{data_source}/dry-plan
{
  "manifestStr": "eyJjYXRhbG9nIjoibXlfY2F0YWxvZyIsInN......",
  "sql": "SELECT orderkey, order_cust_key FROM \"Orders\" LIMIT 1"
}
```
```
WITH
    "Orders" AS (
        SELECT
            "Orders"."orderkey" AS "orderkey",
            "Orders"."custkey" AS "custkey",
            "Orders"."order_cust_key" AS "order_cust_key"
        FROM
            (
                SELECT
                    "Orders"."orderkey" AS "orderkey",
                    "Orders"."custkey" AS "custkey",
                    "Orders"."order_cust_key" AS "order_cust_key"
                FROM
                    (
                        SELECT
                            o_orderkey AS "orderkey",
                            o_custkey AS "custkey",
                            CONCAT (o_orderkey, '_', o_custkey) AS "order_cust_key"
                        FROM
                            (
                                SELECT
                                    o_orderkey,
                                    o_custkey
                                FROM
                                    public.orders
                            ) AS "Orders"
                    ) AS "Orders"
            ) AS "Orders"
    )
SELECT
    orderkey,
    order_cust_key
FROM
    "Orders"
LIMIT
    1
```

### DIFF
--- a/ibis-server/app/mdl/rewriter.py
+++ b/ibis-server/app/mdl/rewriter.py
@@ -1,21 +1,47 @@
 import httpx
 import orjson
+import sqlglot
+from model.data_source import DataSource
 
 from app.config import get_config
-from app.logger import log_rewritten
+from app.logger import get_logger
 
 wren_engine_endpoint = get_config().wren_engine_endpoint
 
+logger = get_logger("app.mdl.rewriter")
 
-@log_rewritten
-def rewrite(manifest_str: str, sql: str) -> str:
-    try:
-        r = httpx.request(
-            method="GET",
-            url=f"{wren_engine_endpoint}/v2/mdl/dry-plan",
-            headers={"Content-Type": "application/json", "Accept": "application/json"},
-            content=orjson.dumps({"manifestStr": manifest_str, "sql": sql}),
-        )
-        return r.text if r.status_code == httpx.codes.OK else r.raise_for_status()
-    except httpx.ConnectError as e:
-        raise ConnectionError(f"Can not connect to Wren Engine: {e}")
+
+class Rewriter:
+    def __init__(self, manifest_str: str, data_source: DataSource = None):
+        self.manifest_str = manifest_str
+        self.data_source = data_source
+
+    def rewrite(self, sql: str) -> str:
+        try:
+            r = httpx.request(
+                method="GET",
+                url=f"{wren_engine_endpoint}/v2/mdl/dry-plan",
+                headers={
+                    "Content-Type": "application/json",
+                    "Accept": "application/json",
+                },
+                content=orjson.dumps({"manifestStr": self.manifest_str, "sql": sql}),
+            )
+            rewritten_sql = (
+                r.text if r.status_code == httpx.codes.OK else r.raise_for_status()
+            )
+            logger.debug(f"Rewritten SQL: {rewritten_sql}")
+            return (
+                rewritten_sql
+                if self.data_source is None
+                else self.transpile(rewritten_sql)
+            )
+        except httpx.ConnectError as e:
+            raise ConnectionError(f"Can not connect to Wren Engine: {e}")
+
+    def transpile(self, rewritten_sql: str) -> str:
+        transpiled_sql = sqlglot.transpile(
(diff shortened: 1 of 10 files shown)


onlyjackfrost:
@grieve54706  Thanks for implementing these two APIs! 
Just one question, 
Is the SQL from the `POST /v2/ibis/dry-plan` API Trino-SQL? while the other is the native SQL that can be executed on each data source.

grieve54706:
> @grieve54706 Thanks for implementing these two APIs! Just one question, Is the SQL from the `POST /v2/ibis/dry-plan` API Trino-SQL? while the other is the native SQL that can be executed on each data source.

The API will return Trino SQL now. But in the future, it will be Datafusion SQL when integrating Rust. They are not for each data source.

ibis-server/app/mdl/rewriter.py
```diff
@@ -1,21 +1,50 @@
...
+        except httpx.ConnectError as e:
+            raise ConnectionError(f"Can not connect to Wren Engine: {e}")
+
+    def transpile(self, rewritten_sql: str) -> str:
+        transpiled_sql = sqlglot.transpile(
+            rewritten_sql, read="trino", write=self.data_source.name
+        )[0]
+        transpiled_sql = sqlglot.parse_one(
+            transpiled_sql, dialect=self.data_source.name
+        ).sql(self.data_source.name)
```
goldmedal, line 48: Curiously, why we need this line? I do some tests, then I found the result of `sqlglot.parse_one().sql()` is equal to `sqlglot.transpile`? There're any additional check in this api?

ibis-server/tests/routers/test_ibis.py
goldmedal, line 1: Why the file called `test_ibis.py`? I think we can call it `test_dry_plan.py`?

goldmedal (approved):
Thanks @grieve54706 worked on this. LGTM. Just have some minor questions.

ibis-server/app/mdl/rewriter.py
grieve54706: Weird, I found a problem if only using `sqlglot.transpile` will get a SQL with invalid syntax `'` like `'WITH ABC ....'`. But now I re-test again, and it is gone.
I will remove this code. Thanks.

ibis-server/tests/routers/test_ibis.py
grieve54706: Make the test include APIs without the data source in `/v2/ibis`.

goldmedal: I see. Thanks

PR #649, opened 2024-07-03
Remove redundent brackets for the join syntax

goldmedal:
# Description
Close #647 
The join relation with brackets is valid for Trino but isn't for Clickhouse.
```
SELECT *
FROM (
  a
  INNER JOIN b ON (a.x = b.y)
)
```

Because `ibis` can't handle the syntax for Clickhouse, we have stopped generating the join syntax with brackets in this PR.
```
SELECT *
FROM
  a
  INNER JOIN b ON (a.x = b.y)
```

# Related Changed
- Currenlty, `dynamic-query` will be enable by default for wren-engine.

PR #658, opened 2024-07-04
Fix the alias join for Non-Clickhouse connector

goldmedal:
# Description
Expect `Clickhouse`, the most data source support the alias join relation. However, we remove the brackets in #649. It will cause some problem when someone use alias join in other connector.

The followed SQL should be valid in non-clickhouse:
```sql
SELECT orderstatus FROM ("Orders" o JOIN "Customer" c ON o.custkey = c.custkey) j1 LIMIT 1
```
It's also hard to implement for sqlglot https://github.com/tobymao/sqlglot/issues/3735#issuecomment-2208028722
We needs to avoid this syntax ASAP.

# Changed
- Fix the SQL analysis tests in 2cd455d

### DIFF
--- a/ibis-server/tests/routers/ibis/test_clickhouse.py
+++ b/ibis-server/tests/routers/ibis/test_clickhouse.py
@@ -349,6 +349,19 @@ class TestClickHouse:
             "totalprice": "object",
         }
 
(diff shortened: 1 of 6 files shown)


grieve54706 (approved):
LGTM

PR #659, opened 2024-07-05
Rename API and deconstruct test class

grieve54706:
- Change API path from `/v2/ibis/{data_source}` to `/v2/connector/{data_source}`
- Deconstruct test class


ibis-server/tests/routers/v2/connector/test_postgres.py
```diff
@@ -0,0 +1,429 @@
...
+import pandas as pd
+import pytest
+import sqlalchemy
+from fastapi.testclient import TestClient
+from testcontainers.postgres import PostgresContainer
+
+from app.main import app
+from tests.confest import file_path
+
+# pytestmark = pytest.mark.postgres
```
goldmedal, line 13: Is it correct? I guess it should not be commented?
```suggestion
pytestmark = pytest.mark.postgres
```

goldmedal (approved):
Thanks @grieve54706, overall LGTM but one comment.

PR #668, opened 2024-07-12
Make trino test independent

grieve54706:
The Trino test container is unstable, causing occasional test failures. We have split the tests until the issue is resolved.


goldmedal (approved):
Many Thanks @grieve54706 LGTM

PR #670, opened 2024-07-15
Make the fixture wait until TPCH is ready

grieve54706:


grieve54706:
We have no way to fix it. https://github.com/Canner/wren-engine/issues/671

PR #672, opened 2024-07-16
Mask sensitive information

grieve54706:
Mask all fields of connection info.

### DIFF
--- a/ibis-server/app/model/__init__.py
+++ b/ibis-server/app/model/__init__.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
-from pydantic import BaseModel, Field
+from urllib.parse import urlparse
+
+from pydantic import BaseModel, Field, SecretStr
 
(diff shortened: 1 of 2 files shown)


ibis-server/app/model/__init__.py
```diff
@@ -48,23 +50,23 @@ class QueryTrinoDTO(QueryDTO):
...
+    password: SecretStr
 
 
 class MSSqlConnectionInfo(BaseModel):
     host: str
     port: int
     database: str
     user: str
-    password: str
+    password: SecretStr
```
goldmedal, line 69: I think we should mask all info in the connection info. People might not want to expose their host, port, or any information that could make others try to access their machine.

grieve54706: Sure

goldmedal (approved):
Thanks @grieve54706 Looks good to me.

PR #674, opened 2024-07-16
Fix missing field for Trino

grieve54706:
Follow https://github.com/Canner/wren-engine/pull/672


goldmedal (approved):
Thanks @grieve54706 LGTM.

PR #675, opened 2024-07-17
Show original characters for string literals instead of escaping

### LINKED ISSUE #673: MySQL(mariadb): Pattern match a varchar column with a non-English scalar value.
The discord user report this issue: https://discord.com/channels/1227143286951514152/1238608474003345468/1262627269982031872

- It seems the Korean literal is converted to a literal `U&' ... '`. We should check if this behavior is correct.
- The `lower()` function is generated by Wren AI. I'm not sure if it's correct for MariaDB but I think it works in DuckDB
```
D select lower('시험') = '시험';
┌──────────────────────────┐
│ (lower('시험') = '시험') │
│         boolean          │
├──────────────────────────┤
│ true                     │
└──────────────────────────┘
```

grieve54706:
Resolve #673

Follow Trino
https://github.com/trinodb/trino/pull/18256

### DIFF
--- a/trino-parser/src/main/java/io/trino/sql/ExpressionFormatter.java
+++ b/trino-parser/src/main/java/io/trino/sql/ExpressionFormatter.java
@@ -1,6 +1,5 @@
 package io.trino.sql;
 
-import com.google.common.base.CharMatcher;
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
 import io.trino.sql.SqlFormatter.Dialect;
@@ -92,7 +91,6 @@ import java.text.DecimalFormatSymbols;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
-import java.util.PrimitiveIterator;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
@@ -237,13 +235,13 @@ public final class ExpressionFormatter
         @Override
         protected String visitStringLiteral(StringLiteral node, Void context)
         {
-            return formatStringLiteral(node.getValue(), dialect);
+            return formatStringLiteral(node.getValue());
         }
 
         @Override
         protected String visitCharLiteral(CharLiteral node, Void context)
         {
-            return "CHAR " + formatStringLiteral(node.getValue(), dialect);
+            return "CHAR " + formatStringLiteral(node.getValue());
         }
 
         @Override
@@ -327,7 +325,7 @@ public final class ExpressionFormatter
         @Override
         protected String visitGenericLiteral(GenericLiteral node, Void context)
         {
-            return node.getType() + " " + formatStringLiteral(node.getValue(), dialect);
+            return node.getType() + " " + formatStringLiteral(node.getValue());
         }
 
         @Override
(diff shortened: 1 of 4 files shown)


trino-parser/src/test/java/io/trino/sql/parser/TestStatementBuilder.java
```diff
@@ -331,11 +331,14 @@ public void testStatementBuilder()
...
         assertSqlFormatter("'hello world'", "'hello world'");
-        assertSqlFormatter("U&'!+10FFFF!6d4B!8Bd5ABC!6d4B!8Bd5' UESCAPE '!'", "U&'\\+10FFFF\\6D4B\\8BD5ABC\\6D4B\\8BD5'");
-        assertSqlFormatter("U&'\\+10FFFF\\6D4B\\8BD5\\0041\\0042\\0043\\6D4B\\8BD5'", "U&'\\+10FFFF\\6D4B\\8BD5ABC\\6D4B\\8BD5'");
-        assertSqlFormatter("U&'\\\\abc\\6D4B'''", "U&'\\\\abc\\6D4B'''");
+        assertSqlFormatter("U&'!+10FFFF!6d4B!8Bd5ABC!6d4B!8Bd5' UESCAPE '!'", "'\uDBFF\uDFFF测试ABC测试'");
+        assertSqlFormatter("U&'\\+10FFFF\\6D4B\\8BD5\\0041\\0042\\0043\\6D4B\\8BD5'", "'\uDBFF\uDFFF测试ABC测试'");
+        assertSqlFormatter("U&'\\\\abc\\6D4B'''", "'\\abc测'''");
+        assertSqlFormatter("'攻殻機動隊'", "'攻殻機動隊'");
+        assertSqlFormatter("'😂'", "'😂'");
+        assertSqlFormatter("'시험'", "'시험'");
```
goldmedal, line 341: 👍

wren-tests/src/test/java/io/wren/testing/duckdb/TestWrenWithDuckDBTableFunction.java
```diff
@@ -33,6 +33,7 @@
 import static java.util.Objects.requireNonNull;
 import static org.assertj.core.api.Assertions.assertThat;
 
+@Test(singleThreaded = true)
```
goldmedal, line 36: Note: Reloading DuckDB would be triggered by other test cases (set init-sql). Run tests sequentially to avoid testing failure.

trino-parser/src/main/java/io/trino/sql/ExpressionFormatter.java
```diff
@@ -998,61 +996,9 @@ private String processDateDiffInBigQuery(FunctionCall node, Void context)
...
-        }
-        else {
-            s = s.replace("'", "''");
-        }
-        if (CharMatcher.inRange((char) 0x20, (char) 0x7E).matchesAllOf(s)) {
-            return "'" + s + "'";
-        }
-
-        StringBuilder builder = new StringBuilder();
-        if (dialect == DEFAULT || dialect == DUCKDB || dialect == POSTGRES) {
```
goldmedal, line 1014: I saw there're some DuckDB-specific behaviors. Do they work well if we remove the dialect?

wren-tests/src/test/java/io/wren/testing/duckdb/TestWrenWithDuckDBTableFunction.java
grieve54706: Another test class, `TestDuckDBResource`, which uses `setDuckDBInitSQL`, is already set to `singleThreaded`. They don't affect each other. We don't need to make any adjustments now. If we encounter any issues in the future, we can then modify the test suite as necessary.

trino-parser/src/main/java/io/trino/sql/ExpressionFormatter.java
goldmedal: Sounds great.

goldmedal (approved):
Thanks @grieve54706 LGTM.

PR #676, opened 2024-07-17
Allow to invoke a calculation belong to another models

goldmedal:
# Description
This PR allows the calculation to be used by another model through the relationship. Consider the case in `wren-modeling-rs/wren-example/examples/calculation-invoke-calculation.rs`:
```rust
.column(
    ColumnBuilder::new("customer_state_cf", "varchar")
        .calculated(true)
        .expression("orders.customer_state")
        .build(),
)
```
We can add a calculated field in `order_items` to access the calculated field `customer_state` in `orders`.

# Known Issue
- Only allow one calculation in an expression currently. We will file another issue for it.

### DIFF
--- a/wren-modeling-rs/Cargo.toml
+++ b/wren-modeling-rs/Cargo.toml
@@ -14,7 +14,7 @@ version = "0.1.0"
 
 [workspace.dependencies]
 async-trait = "0.1.80"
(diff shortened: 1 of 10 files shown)


grieve54706 (approved):
LGTM. Thanks @goldmedal.

PR #678, opened 2024-07-18
Fix the format of TableReference

### LINKED ISSUE #591: Fix the format of TableReference
# Description
The table reference of a model is a object but it's a string in rust version now.
```
{
   "catalog": "c1",
   "schema": "s1",
   "table": "t1"
}
```
https://github.com/goldmedal/wren-engine/blob/930a8d8bf9e944d4f127b0bad2bab67606df4359/wren-modeling-rs/src/mdl/manifest.rs#L29

The catalog and schema are optional but table should be required.

goldmedal:
close #591

### DIFF
--- a/wren-modeling-py/src/lib.rs
+++ b/wren-modeling-py/src/lib.rs
@@ -52,7 +52,10 @@ mod tests {
             "models": [
                 {
                     "name": "customer",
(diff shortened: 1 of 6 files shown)


grieve54706 (approved):
LGTM. Thanks @goldmedal

PR #681, opened 2024-07-18
Upgrade DataFusion to 40.0.0

goldmedal:
# Description
- Update the API usage.
- Fix the test case.

### DIFF
--- a/wren-modeling-rs/core/src/mdl/mod.rs
+++ b/wren-modeling-rs/core/src/mdl/mod.rs
@@ -264,15 +264,13 @@ mod test {
     use std::path::PathBuf;
     use std::sync::Arc;
 
+    use datafusion::arrow::array::{ArrayRef, Int32Array, RecordBatch, StringArray};
     use datafusion::common::not_impl_err;
-    use datafusion::error::Result;
+    use datafusion::common::Result;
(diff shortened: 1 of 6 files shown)


wren-modeling-rs/core/src/mdl/mod.rs
```diff
@@ -313,9 +310,9 @@ mod test {
                 "select orderkey + orderkey from test.test.orders",
                 "select orderkey from test.test.orders where orders.totalprice > 10",
                 "select orders.orderkey from test.test.orders left join test.test.customer on (orders.custkey = customer.custkey) where orders.totalprice > 10",
-                "select orderkey, sum(totalprice) from test.test.orders group by 1",
+                "select orderkey, min(totalprice) from test.test.orders group by 1",
```
goldmedal, line 313: The `sum` function is moved to `datafusion-function-aggregate` after 40.0.0. The aggregate function crate won't be registered for the custom context provider by default. We should use `SessionContext` in https://github.com/Canner/wren-engine/blob/daf5a60fd4695e85fb152c589c83cd4c5a328f40/wren-modeling-rs/core/src/mdl/utils.rs#L63

I will do it in another PR.

PR #684, opened 2024-07-19
Create expressions by SessionContext API

goldmedal:
# Description
As mentioned in https://github.com/Canner/wren-engine/pull/681#discussion_r1682494258, DataFusion moves many functions out of the core. If we used the context provider and created the expression by ourselves, we should register many functions for it. This PR uses `SessionCotnext#SessionState` to create what we need.

### DIFF
--- a/wren-modeling-rs/core/src/mdl/utils.rs
+++ b/wren-modeling-rs/core/src/mdl/utils.rs
@@ -276,7 +276,7 @@ mod tests {
         let ctx = SessionContext::new();
         let model = analyzed_mdl.wren_mdl().get_model("customer").unwrap();
         let expr = super::create_wren_expr_for_model(
-            &"name".to_string(),
+            "name",
             Arc::clone(&model),
             ctx.state_ref(),
         )?;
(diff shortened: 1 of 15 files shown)


wren-modeling-rs/core/src/mdl/utils.rs
```diff
@@ -89,105 +86,97 @@ pub fn create_wren_calculated_field_expr(
...
-    let statement = &parsed[0];
-
-    let sql_to_rel = SqlToRel::new(&context_provider);
-    let plan = sql_to_rel.sql_statement_to_plan(statement.clone())?;
-    match plan {
-        LogicalPlan::Projection(projection) => Ok(projection.expr[0].clone()),
-        _ => internal_err!("Unexpected plan type: {:?}", plan),
-    }
+/// Add the table prefix for the column expression if it can be resolved by the schema.
+fn qualified_expr(
```
goldmedal, line 157: I have filed an issue https://github.com/apache/datafusion/issues/11551 for it. If the behavior is improved, we can remove this function.

grieve54706 (approved):
LGTM. Thanks @goldmedal.

PR #686, opened 2024-07-23
Use GitHub App instead of PAT

grieve54706:


goldmedal (approved):
LGTM. Let's try it first.

PR #687, opened 2024-07-23
Fix `RemoveWrenPrefixRule` for being executed by DataFusion

goldmedal:
# Description
- Fix the `RemoveWrenPrefixRule` for the correct plan schema.
- Provide an API for applying rules for an existing session context.
- Make sqlloigctests tests to apply rules directly.

## Fix the sqllogictests
I didn't find the sqllogictests is broken. So, many test cases are failed and skipped but I didn't know. I commented out some cases and I'll file the issues for them.

### DIFF
--- a/wren-modeling-rs/core/src/mdl/context.rs
+++ b/wren-modeling-rs/core/src/mdl/context.rs
@@ -19,6 +19,9 @@ use crate::logical_plan::utils::create_schema;
 use crate::mdl::manifest::Model;
 use crate::mdl::{AnalyzedWrenMDL, WrenMDL};
 
+/// Apply Wren Rules to the context for sql generation.
+/// TODO: There're some issue for unparsing the datafusion optimized plans.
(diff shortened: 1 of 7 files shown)


wren-modeling-rs/core/src/mdl/context.rs
```diff
@@ -19,6 +19,9 @@ use crate::logical_plan::utils::create_schema;
 use crate::mdl::manifest::Model;
 use crate::mdl::{AnalyzedWrenMDL, WrenMDL};
 
+/// Apply Wren Rules to the context for sql generation.
+/// TODO: There're some issue for unparsing the datafusion optimized plans.
+///   Disable all the optimize rule for sql generation temporarily.
```
goldmedal, line 24: There's an issue about unparsing the optimized plans to the SQL string. I'll file the issue in Datafusion.

grieve54706 (approved):

PR #689, opened 2024-07-23
chore(trino): upgrade testcontainers to 4.7.2

grieve54706:
Upgrade testcontainers and use the `TrinoContainer` of testcontainers instead of the container we made


ibis-server/app/model/data_source.py
```diff
@@ -163,6 +163,6 @@ def get_trino_connection(
             port=int(info.port.get_secret_value()),
             database=info.catalog.get_secret_value(),
             schema=info.trino_schema.get_secret_value(),
-            user=info.user.get_secret_value(),
-            password=info.password.get_secret_value(),
+            user=(info.user and info.user.get_secret_value()),
+            password=(info.password and info.password.get_secret_value()),
```
goldmedal, line 167: Curiously, why need to have an `and` operation here? For null checking? If yes, should we do the same thing for other properties?

goldmedal (approved):
Thanks @grieve54706 looks nice. Just one question.
BTW, I think you can link this PR to your PR in testcontainers repo. https://github.com/testcontainers/testcontainers-python/pull/642/files

ibis-server/app/model/data_source.py
grieve54706: Yes, for null. Other information about the data source is not allowed to be null.

PR #692, opened 2024-07-24
Decouple the Rewriter and Connector in Ibis Server

### LINKED ISSUE #691: Decouple Rewriter and Connector of Ibis
# Description
We plan to provide the `v3` API for the beta release of Wren Core (Rust version). The main difference between `v3` and `v2` is that the Rewriter will be powered by a different core. The `v2` will use the legacy Wren core (Java version) to plan SQL, while `v3` will use the latest Wren core (Rust version in beta) to plan SQL.

The basic concept is to make the Connector only accept SQL and then query this without planning. In the API layer, we need to invoke a different Rewriter to plan the SQL. It would be like:
```python
def query(
    data_source: DataSource,
    dto: QueryDTO,
    dry_run: Annotated[bool, Query(alias="dryRun")] = False,
    limit: int | None = None,
) -> Response:
    rewritten_sql = Rewriter(self.manifest_str, self.data_source).rewrite(sql)
    connector = Connector(data_source, dto.connection_info, dto.manifest_str)
    if dry_run:
        connector.dry_run(rewritten_sql)
        return Response(status_code=204)
    return JSONResponse(
        to_json(connector.query(rewritten_sql, limit=limit), dto.column_dtypes)
    )
```

goldmedal:
- close #691 
- I also enhanced the testing scripts in justfile and added some corresponding docs.

### DIFF
--- a/ibis-server/README.md
+++ b/ibis-server/README.md
@@ -78,5 +78,17 @@ Run the server
 just run
 ```
 
+### Run the testing
+- Preapre the Wren Engine server (see [Wren Engine Example](../example/README.md)
+- Run pytest with `WREN_ENGINE_ENDPOINT` env. (The default port of Wren Engine is 8080)
+  - run all test
+  ```
+  WREN_ENGINE_ENDPOINT=http://localhost:8080 just test
+  ```
+  - run specific test using [pytest args](https://docs.pytest.org/en/latest/example/markers.html#using-k-expr-to-select-tests-based-on-their-name)
+  ```
+  WREN_ENGINE_ENDPOINT=http://localhost:8080 just test -k 'postgres'
+  ```
+
 ## Contributing
 Please see [CONTRIBUTING.md](docs/CONTRIBUTING.md) for more information.
--- a/ibis-server/README.md
+++ b/ibis-server/README.md
@@ -78,17 +78,5 @@ Run the server
 just run
 ```
 
-### Run the testing
-- Preapre the Wren Engine server (see [Wren Engine Example](../example/README.md)
-- Run pytest with `WREN_ENGINE_ENDPOINT` env. (The default port of Wren Engine is 8080)
-  - run all test
-  ```
-  WREN_ENGINE_ENDPOINT=http://localhost:8080 just test
-  ```
-  - run specific test using [pytest args](https://docs.pytest.org/en/latest/example/markers.html#using-k-expr-to-select-tests-based-on-their-name)
-  ```
-  WREN_ENGINE_ENDPOINT=http://localhost:8080 just test -k 'postgres'
-  ```
-
 ## Contributing
 Please see [CONTRIBUTING.md](docs/CONTRIBUTING.md) for more information.
--- a/ibis-server/docs/development.md
+++ b/ibis-server/docs/development.md
@@ -48,7 +48,7 @@ To start the server:
 ### Run the testing
 - Preapre the Wren Engine server (see [Wren Engine Example](../example/README.md)
 - Run pytest with `WREN_ENGINE_ENDPOINT` env. (The default port of Wren Engine is 8080)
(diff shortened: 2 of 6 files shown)


ibis-server/README.md
```diff
@@ -78,5 +78,17 @@ Run the server
...
+- Run pytest with `WREN_ENGINE_ENDPOINT` env. (The default port of Wren Engine is 8080)
+  - run all test
+  ```
+  WREN_ENGINE_ENDPOINT=http://localhost:8080 just test
+  ```
+  - run specific test using [pytest args](https://docs.pytest.org/en/latest/example/markers.html#using-k-expr-to-select-tests-based-on-their-name)
+  ```
+  WREN_ENGINE_ENDPOINT=http://localhost:8080 just test -k 'postgres'
+  ```
+
```
grieve54706, line 92: Let's move this block into development.md

ibis-server/justfile
```diff
@@ -31,8 +31,9 @@ run:
 dev:
     poetry run fastapi dev --port {{port}}
 
-test:
-    poetry run pytest
+# run the pytest with given args
+test *ARGS:
+    poetry run pytest {{ ARGS }}
```
grieve54706, line 36: I prefer `poetry run pytest -m {{marker}}` to restrict contributors to using markers. If they want other testing methods, they can directly use the `poetry run pytest` command.

goldmedal: I don't prefer this because `marker` would be the required option for running testing. However, if we just append the args to the tail, it's an optional args. `just test` also works for running all tests. It would like an alias of `poetry run pytest`.
If we want to have a command for restricting tests, I guess we can provider another just command for it. Maybe
```
db-test *ARGS:
    poetry run pytest -m {{ ARGS[0] }}
```
I'm not sure.

grieve54706: We could only provide one command for testing the specific marker. We now have too many test suites requiring various Docker images, and the chances of needing to run all tests together are quite low. If there's a genuine need to run all tests together, one can execute `poetry run pytest` directly.

ibis-server/docs/development.md
```diff
@@ -48,6 +48,17 @@ To run the tests:
...
+- Preapre the Wren Engine server (see [Wren Engine Example](../example/README.md)
+- Run pytest with `WREN_ENGINE_ENDPOINT` env. (The default port of Wren Engine is 8080)
+  - run all test
+  ```
+  WREN_ENGINE_ENDPOINT=http://localhost:8080 just test
+  ```
+  - run specific test using [pytest args](https://docs.pytest.org/en/latest/example/markers.html#using-k-expr-to-select-tests-based-on-their-name)
+  ```
+  WREN_ENGINE_ENDPOINT=http://localhost:8080 just test -m 'postgres'
+  ```
```
grieve54706, line 61: Could you move the `To run the tests` at L#40 into this block?

goldmedal: I guess we can remove `To run the tests` because it's duplicated with this block.

grieve54706: Sure

ibis-server/docs/development.md
```diff
@@ -48,6 +45,13 @@ To run the tests:
 - Build the image: `just docker-build`
 - Run the container: `just docker-run`
 
+### Run the testing
+- Preapre the Wren Engine server (see [Wren Engine Example](../example/README.md)
+- Run pytest with `WREN_ENGINE_ENDPOINT` env. (The default port of Wren Engine is 8080)
+  - run specific data source test using [pytest marker](https://docs.pytest.org/en/stable/example/markers.html). There are some markers for different data sources. See the list in [pyproject.toml](https://github.com/Canner/wren-engine/blob/10d71be41ecb630e6d5f56834d2724afbf87710d/ibis-server/pyproject.toml#L50).
```
grieve54706, line 51: I'm sorry late to notice this `pyproject.toml` link. How about using `../pyproject.toml`? If we use like with commit sha and line number, it will not follow the latest.

grieve54706 (approved):
LGTM, just minor.

PR #693, opened 2024-07-24
chore(dependabot): add `dependabot.yml` for multi module

grieve54706:
Reference: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file


goldmedal (approved):
Thanks @grieve54706 It's very helpful. LGTM.

PR #706, opened 2024-07-24
chore(dependabot): group dependencies

grieve54706:
Group dependencies by module and separate two python module


goldmedal (approved):
LGTM.

PR #713, opened 2024-07-24
chore(log): use loguru instead of logging

grieve54706:
### Use [loguru](https://github.com/Delgan/loguru) instead of logging.
It could provide more features, such as associating request logs with a UUID, which is particularly helpful in multi-threaded environments.
Additionally, it allows the entire stack trace to be displayed, including the values of variables.

### Add API to control diagnose
Loguru allows log values of variables when `diagnose` is True. Ref [fully-descriptive-exceptions](https://github.com/Delgan/loguru?tab=readme-ov-file#fully-descriptive-exceptions)

### Remove log level
The main principle of `loguru` is that there are no levels. By default, everything is printed to `sys.stderr`. It then allows you to choose to output specific levels to designated files. Please refer to https://github.com/Delgan/loguru/issues/138#issuecomment-525844687

### Use middleware instead of log_dto
By using [middleware](https://fastapi.tiangolo.com/tutorial/middleware/), we can more elegantly print out the request's parameters and body, and have their logs associated with a UUID. But the request body is not deserialized to the Pydantic model, we need to deserialize the body to JSON and then mask connection information.

### Pass Java engine response
Previously, we would only return the `Server Internal Error` message that occurred when ibis-server called the Java engine to the caller. Now, we will return the error message from the Java engine itself.

### Rename variable
Make the rewritten sql with the manifest name `planned_sql`.
Make the transpired sql by sqlglot name `dialect_sql`.

### Minor
- Fix config not stand-alone
- Simply use raise_for_status
- Add UnprocessableEntityError


ibis-server/docs/development.md
```diff
@@ -42,7 +42,6 @@ To run the tests:
 
 ### Environment Variables
 - `WREN_ENGINE_ENDPOINT`: The endpoint of the Wren Java engine
-- `LOG_LEVEL`: The log level of the server (default is INFO)
```
goldmedal, line 45: Curiously, it seems that `loguru` doesn't use this env. How can we change the log level after moving to `loguru`?

goldmedal (commented):
Thanks @grieve54706 there're some conflicts here.

ibis-server/docs/development.md
grieve54706: The main principle of `loguru` is that there are no levels. By default, everything is printed to `sys.stderr`. It then allows you to choose to output specific levels to designated files. Please refer to https://github.com/Delgan/loguru/issues/138#issuecomment-525844687

goldmedal: So, everything is logged to `sys.stderr`. IMO, `sys.stderr` is not always printed in the console. Maybe we should check whether the environment in the docker container will print it or not.

ibis-server/app/mdl/rewriter.py
```diff
@@ -46,17 +45,17 @@ def rewrite(self, sql: str) -> str:
...
+            logger.debug("Rewritten SQL: {}", rewritten_sql)
             return (
                 rewritten_sql
                 if self.data_source is None
                 else self.transpile(rewritten_sql)
             )
         except httpx.ConnectError as e:
             raise ConnectionError(f"Can not connect to Wren Engine: {e}")
+        except httpx.HTTPStatusError as e:
+            raise RewriteError(e.response.text)
```
goldmedal, line 58: I think we can also wrap the rewrite exception into RewriteError for EmbeddedEngineRewriter, then I guess we can be powered by the exception handler in `v3/connector.py` to generate the response.

ibis-server/app/main.py
```diff
@@ -23,20 +39,24 @@ def health():
...
 
 
-@app.exception_handler(QueryDryRunError)
-async def query_dry_run_error_handler(request, exc: QueryDryRunError):
-    return PlainTextResponse(str(exc), status_code=422)
+@app.patch("/config")
+def update_config(diagnose: bool):
+    config = get_config()
+    config.update(diagnose=diagnose)
+    return config
```
goldmedal, line 50: How about adding a test to check if this update works?

ibis-server/docs/development.md
grieve54706: After testing, the log will be printed in the console.

ibis-server/app/main.py
```diff
@@ -23,20 +49,22 @@ def health():
...
-def config():
+def provide_config():
     return get_config()
 
 
-@app.exception_handler(QueryDryRunError)
-async def query_dry_run_error_handler(request, exc: QueryDryRunError):
-    return PlainTextResponse(str(exc), status_code=422)
+@app.patch("/config")
+def update_config(config_model: ConfigModel):
```
goldmedal, line 57: I saw you `get_config().init_logger` at L17. Should we init the logger again after updating the config?

ibis-server/app/mdl/rewriter.py
```diff
@@ -23,12 +22,12 @@ def __init__(self, manifest_str: str, data_source: DataSource = None):
     def rewrite(self, sql: str) -> str:
         pass
 
-    def transpile(self, rewritten_sql: str) -> str:
-        transpiled_sql = sqlglot.transpile(
-            rewritten_sql, read="trino", write=self.data_source.name
+    def transpile(self, planned_sql: str) -> str:
+        dialect_sql = sqlglot.transpile(
+            planned_sql, read="trino", write=self.data_source.name
```
goldmedal, line 27: 👍

goldmedal (approved):
Thanks @grieve54706 Looks nice to me. Just one question about config updating.

ibis-server/app/main.py
grieve54706: We will remove the old logger and add a new logger when the config is updated in the `config.py`.

PR #727, opened 2024-08-01
Refactor the model planner and fix view planning

### LINKED ISSUE #680: Allow save query as view without column alias
# Description
We have introduced the view in https://github.com/Canner/wren-engine/pull/645 but there are some column limitations for it. Currently, we query the complex expression with the alias name in the view. The followed SQL isn't valid:
```sql
select order_id, sum(price) from wrenai.public.order_items group by order_id
```
It should be
```sql
select order_id, sum(price) as sum_price from wrenai.public.order_items group by order_id
```

There're some todo case that should be enabled:
https://github.com/Canner/wren-engine/blob/daf5a60fd4695e85fb152c589c83cd4c5a328f40/wren-modeling-rs/sqllogictest/src/test_context.rs#L261-L262

goldmedal:
Close #680 

## Remain issues
- https://github.com/Canner/wren-engine/issues/728

### DIFF
--- a/ibis-server/tests/routers/v3/connector/test_postgres.py
+++ b/ibis-server/tests/routers/v3/connector/test_postgres.py
@@ -399,7 +399,7 @@ def test_dry_plan():
     assert response.status_code == 200
     assert (
         response.text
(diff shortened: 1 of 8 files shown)


grieve54706 (approved):
LGTM. Thanks @goldmedal.

PR #729, opened 2024-08-05
Fix the scope analysis of the CTEs

goldmedal:
# Describe
We didn't build the correct scope for the CTE. This leads to some analysis errors when making the decision point analysis.

### DIFF
--- a/wren-base/src/test/java/io/wren/base/sqlrewrite/analyzer/TestDecisionPointAnalyzer.java
+++ b/wren-base/src/test/java/io/wren/base/sqlrewrite/analyzer/TestDecisionPointAnalyzer.java
@@ -676,6 +676,58 @@ public class TestDecisionPointAnalyzer
                     new ExprSource("t1.custkey", "customer", "custkey", new NodeLocation(2, 33)),
                     new ExprSource("t2.custkey", "orders", "custkey", new NodeLocation(2, 50))));
         }
+
+        statement = parseSql("""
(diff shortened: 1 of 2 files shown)


wren-base/src/test/java/io/wren/base/sqlrewrite/analyzer/TestDecisionPointAnalyzer.java
```diff
@@ -676,6 +676,58 @@ WITH t1 as (SELECT * FROM customer), t2 as (SELECT * FROM orders)
...
+        assertThat(result.size()).isEqualTo(3);
+        queryAnalysis = result.get(2);
+        assertThat(queryAnalysis.getSelectItems().get(0).getExprSources().size()).isEqualTo(1);
+
+
+        // we only analyze the top-level expression
+        statement = parseSql("""
+                WITH t1 as (SELECT customer.custkey FROM customer),
+                t2 as (SELECT (t1.custkey + 1) as custkey_plus FROM t1)
+                SELECT custkey_plus FROM t2
```
goldmedal, line 725: We only analyze the top-level expression. We won't dig into the child of this expression. So, the number of expression sources of `custkey_plus` is zero.

cyyeh:
@goldmedal thanks for your help! the issue seems fixed after some manual testing

grieve54706 (approved):
LGTM. Thanks @goldmedal.

PR #734, opened 2024-08-06
fix(mssql): import custom dialects from ibis library for sqlglot

grieve54706:
Resolve https://github.com/Canner/WrenAI/issues/581

Ibis implements the MSSQL dialect for sqlglot. If we use sqlglot before using Ibis, the MSSQL dialect will not be registered in sqlglot. Therefore, we need to import the dialect from Ibis before using sqlglot.


goldmedal (approved):
Thanks @grieve54706  LGTM.

PR #737, opened 2024-08-07
fix(core): fix query a view contain the alias expressions

### LINKED ISSUE #728: Fix the statement of View contains alias column can't be query
# Description
- selecting all works but can't query the specific column.
- Disabled Tests
https://github.com/Canner/wren-engine/blob/6f0957ca47ada675efb6316e665037c2419b257d/wren-modeling-rs/sqllogictest/test_sql_files/view.slt#L4-L7

goldmedal:
# Description
Close #728


grieve54706 (approved):
LGTM. BTW, how about using `core` be scope and let the first letter of the message lowercase?

PR #739, opened 2024-08-08
test(ibis): add supporting functions test for BigQuery data source

goldmedal:
# Description
We need to have a white list of functions for each data source. In this PR, we add function tests for BigQuery according to [trino functions list](https://trino.io/docs/current/functions/list-by-topic.html) to check what we can use.

In this PR, I commanded out the unsupported functions. They can't handled well by sqlglot.

We won't merge this PR because there're some issues about formatting trino SQL and the tests aren't qualified.

# Test Result - Function White List
### Date and Time Functions
- `current_date`
- `current_timestamp`
- `date`
- `date_add`
- `date_diff`
- `date_format`
- `date_parse`
- `date_trunc`
- `from_unixtime`
- `now`

### Aggregate Functions
- `any_value`
- `approx_distinct`
- `arbitrary`
- `array_agg`
- `avg`
- `bool_and`
- `bool_or`
- `count`
- `count_if`
- `max`
- `max_by`
- `min`
- `min_by`
- `sum`

### Math Functions
- `greatest`
- `least`
- `abs`
- `acos`
- `asin`
- `atan`
- `atan2`
- `cbrt`
- `ceil`
- `cos`
- `cosh`
- `exp`
- `floor`
- `is_nan`
- `ln`
- `log`
- `log10`
- `mod`
- `power`
- `round`
- `sign`
- `sin`
- `sinh`
- `sqrt`
- `tan`
- `tanh`

### String Functions
- `chr`
- `concat`
- `format`
- `length`
- `lower`
- `lpad`
- `ltrim`
- `replace`
- `reverse`
- `rpad`
- `rtrim`
- `soundex`
- `split`
- `substr`
- `substring`
- `translate`
- `trim`
- `upper`

### Operators
- Arithmetic Operators: `+`, `-`, `*`, `/`, `%`
- Array Access Operator: `[]`
- Concatenation Operator: `||`
- Comparison Operators: `<`, `>`, `<=`, `>=`, `=`, `<>`, `!=`


trino-parser/src/main/java/io/trino/sql/ExpressionFormatter.java
```diff
@@ -153,7 +153,7 @@ protected String visitNode(Node node, Void context)
         @Override
         protected String visitRow(Row node, Void context)
         {
-            String rowPrefix = (dialect == DEFAULT || dialect == BIGQUERY) ? "ROW" : "";
+            String rowPrefix = "";
```
goldmedal, line 156: There are some issues with formatting Trino SQL. Consider the followed case:
```sql
select sum(c1) from (values (1,2), (2,3)) t(c1, c2)
```
This SQL will be formatted to 
```sql
select sum(c1) from (values ROW(1,2), ROW(2,3)) t(c1, c2)
```
It's valid for Trino but can't be handled correctly when transpiring to the BigQuery sql by sqlglot.
The transpiring result would like
```sql
SELECT SUM(`c1`) FROM UNNEST([STRUCT(STRUCT(1, 2) AS `c1`), STRUCT(STRUCT(2, 3) AS `c1`)])
```
It's wrong. It should be
```sql
SELECT SUM(`c1`) FROM UNNEST([STRUCT(1 AS `c1`, 2 AS `c2`), STRUCT(2 AS `c1`, 3 AS `c2`)])
```

grieve54706:
Thanks @goldmedal. Would you mind adding comments or descriptions for code that are commented out?

PR #740, opened 2024-08-08
fix(core): exclude relationship column and calculated field when selecting all in the analysis result

### LINKED ISSUE #738: selectItems contain relationship type column results in sql analysis api
in the result of sql analysis api, there should be no selectItems having relationship type column

goldmedal:
# Description
close #738 
In the normal query, both `relationship` and `calculated field` won't be selected when selecting all from a model. We should also exclude them in the analysis result.


wren-base/src/test/java/io/wren/base/sqlrewrite/analyzer/TestDecisionPointAnalyzer.java
```diff
@@ -185,6 +190,20 @@ public void testSelectItem()
         assertThat(result.size()).isEqualTo(1);
         assertThat(result.get(0).getSelectItems().size()).isEqualTo(1);
         assertThat(result.get(0).getSelectItems().get(0).getExpression()).isEqualTo("c.*");
+
+        statement = parseSql("SELECT customer_name FROM orders");
+        result = DecisionPointAnalyzer.analyze(statement, DEFAULT_SESSION_CONTEXT, mdl);
+        assertThat(result.size()).isEqualTo(1);
+        assertThat(result.get(0).getSelectItems().size()).isEqualTo(1);
+        assertThat(result.get(0).getSelectItems().get(0).getExpression()).isEqualTo("customer_name");
+        assertThat(result.get(0).getSelectItems().get(0).getExprSources()).isEqualTo(List.of(new ExprSource("customer_name", "orders", "customer_name", new NodeLocation(1, 8))));
```
goldmedal, line 199: We won't dig into the expression of a calculated field. So, in the `exprSource`, we only provide itself.

wren-base/src/test/java/io/wren/base/sqlrewrite/analyzer/TestDecisionPointAnalyzer.java
```diff
@@ -185,6 +190,20 @@ public void testSelectItem()
...
+
+        statement = parseSql("SELECT customer FROM orders");
```
goldmedal, line 201: A relationship column can't be selected directly. we won't fail the case to make the query stable. However, it's an invalid query for Wren engine.

grieve54706 (approved):
LGTM. Thanks @goldmedal.

grieve54706:
Hi @goldmedal, is this pr resolve #738? If it's, how about connecting the issue?

goldmedal:
> Hi @goldmedal, does this pr resolve #738? If it is, would you be able to connect the issue?

Thanks.  I have linked the issues.

PR #742, opened 2024-08-12
fix(mssql): use msodbcsql instead of tdsodbc

grieve54706:
## Description
We found the `tdsodbc` version in the Debian 10 is 1.0 and may be too old not to support the `Azure SQL database`.
We will use `Microsoft ODBC 18 driver` instead of tdsodbc.
It will support SQL server version `2014` or later.

The msodbcsql driver 18.0 supports 2012, but the version is not on `ARM`.
We will use the latest MS ODBC 18 driver for arm64 and amd64. The latest version does not support 2012.

## Additional information
[Operating system support](https://learn.microsoft.com/en-us/sql/connect/odbc/linux-mac/system-requirements?view=sql-server-ver16#operating-system-support)
[SQL version compatibility](https://learn.microsoft.com/en-us/sql/connect/odbc/windows/system-requirements-installation-and-driver-files?view=sql-server-ver16#sql-version-compatibility)

### DIFF
--- a/.github/workflows/ibis-ci.yml
+++ b/.github/workflows/ibis-ci.yml
@@ -54,16 +54,13 @@ jobs:
             ~/.cargo/git/db/
             wren-modeling-py/target/
           key: ${{ runner.os }}-cargo-${{ hashFiles('wren-modeling-py/Cargo.lock') }}
-      - name: Install FreeTDS to be a ODBC driver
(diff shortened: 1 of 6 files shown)


goldmedal (approved):
Thanks @grieve54706 it looks nice to me.

PR #746, opened 2024-08-13
fix(ibis): determine the data is null or not

grieve54706:
If the data is null when the column type is a timestamp, data in the pandas dataframe will be a `NaT`(Not a Time) object. It could not be serialized to JSON in `orjson` ([issue](https://github.com/ijl/orjson/issues/249), [test case]( https://github.com/ijl/orjson/blob/master/test/test_numpy.py#L799)).

NaT is a type of `numpy`. The numpy provides a function [isnat](https://numpy.org/doc/stable/reference/generated/numpy.isnat.html) to determine it. But it will raise an error when the input is `pandas.NaT`. Fortunately, pandas provide a function [isnull](https://pandas.pydata.org/docs/reference/api/pandas.isnull.html) that could cover this type and also cover `NaN`.

### DIFF
--- a/ibis-server/tests/routers/v2/connector/test_bigquery.py
+++ b/ibis-server/tests/routers/v2/connector/test_bigquery.py
@@ -55,6 +55,11 @@ manifest = {
                     "expression": "cast('2024-01-01T23:59:59' as timestamp with time zone)",
                     "type": "timestamp",
                 },
+                {
+                    "name": "test_null_time",
+                    "expression": "cast(NULL as timestamp)",
+                    "type": "timestamp",
+                },
             ],
             "primaryKey": "orderkey",
         },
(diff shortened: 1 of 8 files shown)


ibis-server/tests/routers/v2/connector/test_bigquery.py
```diff
@@ -55,6 +55,11 @@
                     "expression": "cast('2024-01-01T23:59:59' as timestamp with time zone)",
                     "type": "timestamp",
                 },
+                {
+                    "name": "test_null_time",
+                    "expression": "cast(NULL as timestamp)",
+                    "type": "timestamp",
+                },
```
goldmedal, line 62: 👍

goldmedal (approved):
Thanks @grieve54706, overall looks good to me. Just one question, I saw this PR add test for `NaT` type. Do we have any test for `NaN`?

goldmedal:
> Thanks @grieve54706, overall looks good to me. Just one question, I saw this PR add test for `NaT` type. Do we have any test for `NaN`?

After some tests with @grieve54706 , we don't find any way to create a `NaN` case. Let's ignore it here.

PR #749, opened 2024-08-15
chore(test): remove assert text content for dry_plan

grieve54706:
We don't need to assert the content. It will be asserted by the other module.


goldmedal (approved):
👍

PR #754, opened 2024-08-15
ci(integration-tests): add more activity types

grieve54706:
We make the PR from `dependabot` with the label `v1-engine-changed`, but the PR does not trigger ci `integration-test`.
The CI needs more activity types for the PR created with the label.


goldmedal (approved):
👍

PR #755, opened 2024-08-19
feat(core): support case sensitive identifier

### LINKED ISSUE #717: Support the case-sensitive table name and column name
# Description
The model name and column name are currently case insensitive. The following model isn't valid now.
```rust
ModelBuilder::new("Customer")
    .table_reference("datafusion.public.customers")
    .column(ColumnBuilder::new("city", "varchar").build())
    .column(ColumnBuilder::new("id", "varchar").build())
    .column(ColumnBuilder::new("state", "varchar").build())
    .primary_key("id")
    .build(),
```
We don't support the name with the uppercase text.

goldmedal:
# Description
close #717 
After this changed, all name in the MDL will be case sensitive. Because the SQL text in the data fusion is lowercase by default, if the table or column name contains the uppercase, we should use it with quotes.

# Example
```rust
  let manifest = ManifestBuilder::new()
      .catalog("wrenai")
      .schema("public")
      .model(
          ModelBuilder::new("Customer")
              .table_reference("datafusion.public.customer")
              .column(ColumnBuilder::new("custkey", "int").build())
              .column(ColumnBuilder::new("name", "string").build())
              .build(),
      )
      .build();
```

When query the model `Customer`, we should use the following SQL
```sql
SELECT * FROM wrenai.public."Customer"
```

However, the column `custkey` is all lowercase. We can just query it without qutoes:
```sql
SELECT custeky FROM wrenai.public."Customer"
```

### DIFF
--- a/wren-modeling-rs/core/src/mdl/mod.rs
+++ b/wren-modeling-rs/core/src/mdl/mod.rs
@@ -190,6 +190,8 @@ pub fn transform_sql(analyzed_mdl: Arc<AnalyzedWrenMDL>, sql: &str) -> Result<St
 }
 
 /// Transform the SQL based on the MDL with the SessionContext
+/// Wren engine will normalize the SQL to the lower case to solve the case sensitive
+/// issue for the Wren view
 pub async fn transform_sql_with_ctx(
     ctx: &SessionContext,
     analyzed_mdl: Arc<AnalyzedWrenMDL>,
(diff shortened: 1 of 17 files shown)


wren-modeling-rs/core/src/mdl/mod.rs
```diff
@@ -211,7 +213,12 @@ pub async fn transform_sql_with_ctx(
     match plan_to_sql(&analyzed) {
         Ok(sql) => {
             // TODO: workaround to remove unnecessary catalog and schema of mdl
-            let replaced = sql.to_string().replace(&catalog_schema, "");
+            let replaced = sql
+                .to_string()
+                .replace(&catalog_schema, "")
+                // TODO: There're some issue about datafusion unparsing the expr column name with different case
+                // The workaround is to normalize the SQL to the lower case to support Wren View.
+                .to_lowercase();
```
goldmedal, line 221: I'll file the corresponding issue to Datafusion. It generate an invalid SQL for view like
```sql
SELECT 
  Revenue_orders.Order_id, 
  Revenue_orders."sum(Order_items.Price)"  # It's generated from the aggregation in the subquery
FROM 
  (
    SELECT 
      Order_items.Order_id, 
      sum(Order_items.Price) 
    FROM 
      (
        SELECT 
          Order_items.Order_id, 
          Order_items.Price 
        FROM 
          (
            SELECT 
              datafusion.public.order_items.order_id AS Order_id, 
              datafusion.public.order_items.price AS Price 
            FROM 
              datafusion.public.order_items
          ) AS Order_items
      ) AS Order_items 
    GROUP BY 
      Order_items.Order_id
  ) AS Revenue_orders
```

grieve54706 (approved):
LGTM. Thanks @goldmedal.

PR #756, opened 2024-08-19
fix(core): the `properties` should be a json object in the MDL file

### LINKED ISSUE #719: Fix the format of `properties` in the MDL
# Description
The JSON serialized from  the MDL now looks like
```json
        {
          "name": "city",
          "type": "varchar",
          "relationship": null,
          "isCalculated": false,
          "noNull": false,
          "expression": null,
          "properties": []
        },
```
The format of `properties` is wrong. It should be an object.
```
        {
          "name": "city",
          "type": "varchar",
          "relationship": null,
          "isCalculated": false,
          "noNull": false,
          "expression": null,
          "properties":  {
               "key": "value"
                ...
           }
        },
```

goldmedal:
# Description
closed #719


grieve54706 (approved):
LGTM. Thanks @goldmedal.

PR #760, opened 2024-08-20
feat(core): handle JoinType serde using `SCREAMING_SNAKE_CASE` and `snake_case` style

### LINKED ISSUE #718: The JoinType properties of a relationship in MDL should be case insensitive
# Description
I found the join type should now be set in lowercase. 
```
    {
      "name": "orders_customer",
      "models": [
        "orders",
        "customers"
      ],
      "joinType": "many_to_one",
      "condition": "orders.customer_id = customers.id",
      "properties": []
    },
```
However, it should actually be case insensitive.

goldmedal:
# Description
close #718 
Refer to [serde doc](https://serde.rs/container-attrs.html#rename_all), we use `SCREMING_SNAKE_CASE` and add alias for `snake_case`.
The Join type is supported serialized from both of the styles now.
```
one_to_one,  ONE_TO_ONE
```

### DIFF
--- a/wren-modeling-rs/core/src/mdl/builder.rs
+++ b/wren-modeling-rs/core/src/mdl/builder.rs
@@ -430,8 +430,7 @@ mod test {
     }
 
     #[test]
-    fn test_join_type_case_insensitive()
-    {
+    fn test_join_type_case_insensitive() {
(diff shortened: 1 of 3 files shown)


grieve54706 (approved):
LGTM. Thanks @goldmedal.

PR #761, opened 2024-08-20
fix: keep using tdsodbc until web could support

grieve54706:
Revert "fix(mssql): use msodbcsql instead of tdsodbc (#742)"

This reverts commit 50401d1455faf4a992b6f2753a324a3be746bbf9.


goldmedal (approved):
Let's revert it first. Wait the Wren AI team supports it.

PR #762, opened 2024-08-20
feat(core): support the default catalog and schema for wren-core Rust

### LINKED ISSUE #715: Allow to set default catalog and default schema when `transform_sql`
# Description
Currently, `wren-core-rs` recognizes a qualified table name to generate the corresponding model. The user needs to provide the full name of the model in the query, such as:
```sql
SELECT * FROM wren.public.orders
```
It would be more convenient if the user could simply query with the table name:
```sql
SELECT * FROM orders
```

# Solution
We can modify the `transform_sql` API to apply the catalog and schema of MDL to the DataFusion SessionContext.

goldmedal:
closed #715


grieve54706 (approved):
LGMT. Thanks @goldmedal.

PR #766, opened 2024-08-23
feat(core): add TPC-H test and fix the syntax error

goldmedal:
# Description
- Add TPC-H test in sqllogictests
- Disable some unsupported tests and raise the issue.

# Known Issues
- Arrow parses interval literal error.
```
DataFusion error: Arrow error: Parser error: Invalid input syntax for type interval: "0 YEARS 3 MONS 0 DAYS 0 HOURS 0 MINS 0.000000000 SECS seconds"
```
- The schema of the subquery expression is wrong.
```
# TODO:
# External error: query failed: DataFusion error: ModelAnalyzeRule
#  caused by
#  Schema error: No field named partsupp.ps_suppkey.
```
- Calculation column in the view will fail
```
# TODO: fix the calculation column name of view
# statement ok
# SELECT * FROM wrenai.public."Revenue_orders"
```

PR #767, opened 2024-08-23
fix(core): fix the calculation column of the view with case sensitive name

goldmedal:
# Description
Fix the known issue about view of #766


grieve54706 (approved):
LGTM. Thanks @goldmedal.

PR #775, opened 2024-08-30
feat(ibis): adjust json data type and format

grieve54706:
We abandoned the unused feature `column_dtypes`.
If the dtype is datetime, we will format the data to string when building JSON.


grieve54706:
The CI failed cause is from Trino container `GENERIC_INTERNAL_ERROR, message="nodes is empty"`. We can ignore this issue.

goldmedal (approved):
Thanks @grieve54706 Overall LGTM.

PR #776, opened 2024-09-02
fix(core): fix the behaviors of dynamic query

goldmedal:
# Description
- Fix `count(*)` query with the filter that invokes any calculated field.
- Fix `count(*)` for a query with the relation `table` or `join`.
- Fix selecting wildcard from a model should exclude the calculated fields.


grieve54706 (approved):
LGTM

PR #783, opened 2024-09-06
fix(format): handle bytes to hex

grieve54706:
- Handle bytes to hex string
- Override MySQL sqlglot dialect for fixing
- Remove unused table in the test


goldmedal (approved):
Thanks @grieve54706. It looks nice. I think it's good to own our custom dialect

PR #803, opened 2024-09-25
chore(core): remove unused dependency

goldmedal:
# Description
- Remove the unnecessary feature from DataFusion dependency. Take expression-related features only. 
- Change the building profile to `release`

In my local environment, before removing them, the building time is
```
    Finished release profile [optimized] target(s) in 8m 43s
```
After removing them, the time is
```
    Finished release profile [optimized] target(s) in 7m 30s
```

### DIFF
--- a/wren-modeling-py/pyproject.toml
+++ b/wren-modeling-py/pyproject.toml
@@ -17,6 +17,7 @@ pytest = "8.3.3"
 
 [tool.maturin]
 module-name = "wren_core"
+profile = "release"
 include = [{ path = "Cargo.lock", format = "sdist" }]
 exclude = ["tests/**", "target/**"]
 locked = true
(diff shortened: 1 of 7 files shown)


wren-modeling-py/pyproject.toml
```diff
@@ -17,6 +17,7 @@ pytest = "8.3.3"
 
 [tool.maturin]
 module-name = "wren_core"
+profile = "release"
```
grieve54706, line 20: Curious, what does this do?

grieve54706 (approved):
LGTM

wren-modeling-py/pyproject.toml
goldmedal: The `release` profile will optimize the binary more.

PR #805, opened 2024-09-26
chore(core): introduce the benchmark framework for Wren core Rust

goldmedal:
# Description
This PR follows how DataFusion implements [their benchmark](https://github.com/apache/datafusion/tree/main/benchmarks). For Wren core, we only measure the time to plan SQL. We won't actually execute the query.

See more detail in README.md

## Supported Benchmark
- TPC-H


grieve54706 (approved):
LGTM. Thanks @goldmedal.

PR #814, opened 2024-10-03
chore(ibis): remove workaround since ibis update

grieve54706:
Due to certain issues in previous versions of Ibis, we implemented workarounds. Now that Ibis has fixed these issues, let's remove these workarounds.


grieve54706:
The CI failed is `TrinoQueryError(type=INTERNAL_ERROR, name=GENERIC_INTERNAL_ERROR, message="nodes is empty")`. We can ignore it.

goldmedal (approved):
👍

PR #819, opened 2024-10-08
feat(core): apply the DataFusion optimization rule

goldmedal:
# Description
- Apply the DataFusion Optimization rule for a better SQL.
- Now, the Wren Analzye Rule can be registered to SessionContext and executed by DataFusion directly.
- Run the sqllogictests with DataFusion directly.
- Modify the building profile. Build release profile only in stable-release flow.

## Known Issue
- After applying the rules, the calculated field and wren view are broken. I'll address them in another PR.


wren-modeling-py/src/lib.rs
```diff
@@ -71,7 +71,8 @@ mod tests {
                 .unwrap();
         assert_eq!(
             transformed_sql,
-            r#"SELECT * FROM (SELECT main.customer.c_custkey AS c_custkey, main.customer.c_name AS c_name FROM main.customer) AS customer"#
+            "SELECT main.customer.c_custkey AS c_custkey, main.customer.c_name AS c_name FROM \
```
goldmedal, line 74: After applying `ExpandWildcardRule`, the select items are expanded. We will apply the type of coercion rules for them in the future.

grieve54706 (approved):
LGTM. Thanks @goldmedal.

PR #820, opened 2024-10-11
fix(core): fix the field resolving for the relationship column

goldmedal:
# Description
The relationship column should be ignored when resolving the corresponding field. Otherwise, the resolved columns will always contain the relationship column.

Use `RelationType.resolveAnyField` to follow the same behavior.


grieve54706 (approved):
LGTM

PR #822, opened 2024-10-14
fix(core): disable `optimize_projections` rule for SQL generation

goldmedal:
# Description
- After #819, we apply the optimization rule to simplify the generated SQL. However, DataFusion can't unparse the optimized plans well. Disable `optimize_projections` for a valid SQL result.


goldmedal:
It would be fixed by https://github.com/apache/datafusion/pull/12896

grieve54706 (approved):
LGTM

PR #823, opened 2024-10-14
feat(canner): implement new data source `canner`

grieve54706:
## Summary
Support new data source [Canner Enterprise](https://cannerdata.com/product). It provides the PG wire protocol, so we can connect it by following Postgres.

## Connection info
- host: str
- port: str
- user: str
- pat: str
- workspace: str
- enable_ssl: bool, If true, will connect with HTTPS

## Additional information
- Canner enterprise does not support `CREATE TEMPORARY VIEW` to get the schema for ibis. We need to build schema by `dry-run`.
- Canner enterprise does not support `dry-run`, so we have to query with a limit of zero
- The table and column comments are stored in another service of Canner Enterprise. We need to query them by `GraphQL`.
- Preparing Canner Enterprise to test is difficult, so we don't test this feature in the CI.


ibis-server/tests/routers/v2/connector/test_canner.py
```diff
@@ -0,0 +1,365 @@
...
+from app.model.validator import rules
+
+"""
+The Canner Enterprise must setup below:
+- A user with PAT
+- A data source with TPCH Tiny
+- A workspace
+- A TPCH table `orders` in the workspace
+- The table `orders` must be with a comment `This is a table comment`
+- The table `orders` must have a column `o_comment` with a comment `This is a comment`
```
goldmedal, line 18:
```suggestion
- The table `orders` must be with a description `This is a table comment`
- The table `orders` must have a column `o_comment` with a description `This is a comment`
```
In the canner side, I think it's called `description`.

ibis-server/tests/routers/v2/connector/test_canner.py
```diff
@@ -0,0 +1,365 @@
...
+"""
+
+pytestmark = pytest.mark.canner
+
+client = TestClient(app)
```
goldmedal, line 23: I think we can use the `with client` clause here, too.

ibis-server/tests/routers/v2/connector/test_canner.py
```diff
@@ -0,0 +1,365 @@
...
+                {"name": "orderkey", "expression": "o_orderkey", "type": "integer"},
+                {"name": "custkey", "expression": "o_custkey", "type": "integer"},
+                {
+                    "name": "orderstatus",
+                    "expression": "o_orderstatus",
+                    "type": "varchar",
+                    "rls": {
+                        "name": "SESSION_STATUS",
+                        "operator": "EQUALS",
+                    },
```
goldmedal, line 52: Unused field.
```suggestion
```

goldmedal (approved):
Thanks @grieve54706 overall looks good to me. Only some minor comments.

PR #830, opened 2024-10-16
test: close test client after testing and make manifest_str be pytest.fixture

grieve54706:


goldmedal (approved):
Thanks @grieve54706 LGTM

PR #831, opened 2024-10-16
ci(labeler): add PR labeler

grieve54706:


goldmedal (approved):
Thanks @grieve54706 it's really nice 👍

PR #836, opened 2024-10-17
chore(log): adjust log

grieve54706:
- remove newline character of sql
- make the timeout with more information
- print exception message
- add correlation ID and process time to the header
  - response headers will be with the new header `X-Correlation-ID` and `X-Process-Time`.
- add more HTTP status code


goldmedal (approved):
Thanks @grieve54706 LGTM 👍

PR #839, opened 2024-10-18
fix(mssql): handle ibis error when sql with invalid syntax

grieve54706:
We found the error `'NoneType' object has no attribute 'lower'`.
This issue arises because Ibis invokes the `lower()` function on a `None` value when using `_get_schema_using_query()`. 
This PR addresses this by implementing error handling to capture this specific scenario and attempt to extract a more informative error message. 
Moreover, we will continue following Ibis issue https://github.com/ibis-project/ibis/issues/10331.


grieve54706:
Ibis handles this issue at https://github.com/ibis-project/ibis/pull/10332. If we have time to wait for the ibis release, we can close this PR.

goldmedal (approved):
👍

goldmedal:
> Ibis handles this issue at [ibis-project/ibis#10332](https://github.com/ibis-project/ibis/pull/10332). If we have time to wait for the ibis release, we can close this PR.

Filed https://github.com/Canner/wren-engine/issues/840 to trace it

PR #844, opened 2024-10-23
feat(core): infer the source column from the manifest

goldmedal:
# Description
Try to infer the metadata of the source table from the manifest if the register tables aren't provided. 
There are some rules to infer the columns of the source table:
- If the column is calculated or has a relationship, it's not a source column.
-  If the column is without expression, it's a source column.
- If the column has an expression, it will try to infer the source column from the expression.
- If the expression is a simple column reference(`identifier` or `compound identifier`), it's the source column name.
- If the expression is complex, it can't be inferred. 

The source columns will be registered with the source table name in the session context.


grieve54706 (approved):
LGTM. Thanks @goldmedal.

PR #847, opened 2024-10-24
feat(core): enhance the manifest serizlization and deserialization 

goldmedal:
# Description
This PR enhances some issues about the MDLserizlization and deserialization.

## Optional Field
Some fields are optional and can be absent in the MDL. For example, in the model definition, the user only needs to provide one of `tableReference`, `refSql`, or `baseObject`. 
```
{
    "name": "orders",
    "tableReference": {
        "table": "orders"
    },
    "refSql": null  // optional
}
```

## Empty expression
In the expression field of the `Column`, it could be an empty string. We should read it as a null value.
````
{
    "name": "o_orderkey",
    "type": "int",
    "expression": ""
}
````

## Read Bool from integer
In the MDL provided by Wren AI, the boolean value could be `0` or `1`. This PR also backward supports it.
```
{
    "name": "o_orderkey",
    "isCalculated": 0
}
```

## Fix the name of `NotNull`
Fix `no_null` -> `not_null`.


grieve54706 (approved):
LGTM. Thanks @goldmedal.

PR #859, opened 2024-10-29
feat(function-list): search file by data source name

grieve54706:
Change the function file path to a folder path and search the file by data source name.
The system will find the CSV file by the data source name in the path from the env var

## Before
Env var
```
REMOTE_FUNCTION_LIST_PATH=/resource/functions.csv
```

Actual path
```
/resource/functions.csv
```

## After
Env var
```
REMOTE_FUNCTION_LIST_PATH=/resource
```

Actual path by data source
```
/resource/postgres.csv
```


ibis-server/app/config.py
```diff
@@ -58,6 +58,13 @@ def update(self, diagnose: bool):
         else:
             self.init_logger()
 
+    def get_remote_function_list_path(self, data_source: str) -> str:
+        return (
+            f"{self.remote_function_list_path}/{data_source}/functions.csv"
```
goldmedal, line 63: maybe change to this pattern? We don't need another folder, I guess.
```suggestion
            f"{self.remote_function_list_path}/{data_source}.csv"
```

grieve54706: Sure

goldmedal (approved):
Thanks @grieve54706 LGTM 👍

PR #861, opened 2024-10-29
ci(docker): use gha cache

grieve54706:
Use the gha cache to reduce building time. I wish it worked.

Reference:
https://docs.docker.com/build/cache/backends/gha/


goldmedal (approved):
It seems an experimental feature but worth to try, I guess.

PR #862, opened 2024-10-29
feat(core): remove properties field from mdl and improve the error message

goldmedal:
# Description
`properties` is never used on the engine side. We can remove it, and then the Wren AI side can use it more flexibly.


grieve54706 (approved):
LGTM. Thanks @goldmedal.

PR #863, opened 2024-10-30
feat(core): introduce the hidden column

goldmedal:
# Description
This PR introduces a new column property, `is_hidden`. A hidden column means this column is defined in the MDL but it can't be used in the query.
The main purpose of a hidden column is to infer the remote source column. Sometimes, we want to apply something on a remote column like, casting, lowercase, or trim. we may write the MDL like
```json
{
    "name": "date_in_int_col",
    "type": "date",
    "expression": "cast(date_in_int_col as date)"
}
```
However, according to the rule of https://github.com/Canner/wren-engine/pull/844, we can get a remote column from a casting expression. So `date_in_int_col` won't be registered as a source column, the column expression isn't valid.

In this case, we should add another column for the column inferring but we don't want to expose this column. So we can invoke the `is_hidden` property
```json
{
    "name": "date_in_int_col",
    "type": "int",
    "is_hidden": true
},
{
    "name": "date_in_int_col_date",
    "type": "date",
    "expression": "cast(date_in_int_col as date)"
}
```

Now, we can infer the remote column is an int column called `date_in_int_col`. We can use it in the expression of another column.

### DIFF
--- a/.github/workflows/ibis-ci.yml
+++ b/.github/workflows/ibis-ci.yml
@@ -52,7 +52,6 @@ jobs:
             ~/.cargo/registry/index/
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
-            wren-core-py/target/
           key: ${{ runner.os }}-cargo-${{ hashFiles('wren-core-py/Cargo.lock') }}
       - name: Install FreeTDS to be a ODBC driver
         run: |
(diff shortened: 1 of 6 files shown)


grieve54706 (approved):
LGTM

grieve54706:
Hey @goldmedal, some conflicts need to be resolved.

PR #864, opened 2024-10-30
chore(ibis): implement mdl validation scripts

goldmedal:
# Description
Introduce a Python script to validate if a Wren MDL is valid locally.

### DIFF
--- a/ibis-server/pyproject.toml
+++ b/ibis-server/pyproject.toml
@@ -62,6 +62,7 @@ markers = [
 [tool.ruff]
 line-length = 88
 target-version = "py311"
+exclude = ["tools/"]
 
 [tool.ruff.lint]
 select = [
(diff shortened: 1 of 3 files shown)


ibis-server/tools/README.md
```diff
@@ -0,0 +1,7 @@
+# Description
+
+This folder contains useful tools and scripts for debugging and validation.
+
+# Tools
+- `mdl_validation.py`: Used to validate a Wren MDL. This script attempts to select all columns in all models.
```
grieve54706, line 6: Append use example may be good.

grieve54706 (approved):
LGTM. Thank you for providing such a helpful tool.

ibis-server/pyproject.toml
```diff
@@ -62,6 +62,7 @@ markers = [
 [tool.ruff]
 line-length = 88
 target-version = "py311"
+exclude = ["tools/"]
```
goldmedal, line 65: I ignored the ruff for the scripts. I think it's not for production purpose. It would be more flexible.

grieve54706: Hmmm. Having it would be a plus, but it's not essential. Go ahead.

goldmedal:
Thanks @grieve54706

PR #865, opened 2024-10-30
feat(function-list): append file into docker image

grieve54706:
- Move the function list file to ibis-server/resources
- Copy the resources into the docker image
- Fix the file path should exist.


goldmedal (approved):
Thanks @grieve54706 LGTM

PR #869, opened 2024-10-31
chore(ibis): introduce a script for query locally

goldmedal:
# Description
This PR implements a simple Python script to execute an SQL locally. Currently, only implement the BigQuery connection in this PR.
Given the required environment variables, we can input an SQL file like
```
poetry run python tools/query_local_run.py < test.sql
```

### DIFF
--- a/ibis-server/tools/README.md
+++ b/ibis-server/tools/README.md
@@ -5,6 +5,23 @@ This folder contains useful tools and scripts for debugging and validation.
 # Tools
 - `mdl_validation.py`: Used to validate a Wren MDL. This script attempts to select all columns in all models.
   - Requires the `wren_core` library. Run `just install-core` and `just install` before using it.
-```
-poetry run python tools/mdl_validation.py mdl.json function_list/bigquery.csv
-```
+  - Example
+    ```
+    poetry run python tools/mdl_validation.py mdl.json function_list/bigquery.csv
+    ```
+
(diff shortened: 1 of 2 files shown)


grieve54706 (commented):
How about using an env var `WREN_CONFIG_ROOT` and using fixed file names like `manifest.json`, `function_list`, `connection_info.json`?

goldmedal:
> How about using an env var `WREN_CONFIG_ROOT` and using fixed file names like `manifest.json`, `function_list`, `connection_info.json`?

In my use case, I won't put them in the same folder. I could get the manifest from another project and the function list from the ibis resource. I think it's more flexible.

goldmedal:
Because I also use `dotenv` in this script, the user can maintain all of themin the `.env` file.

grieve54706:
How about a `config.properties` file to collect these variables? There are too many environment variables for me now.

goldmedal:
> How about a `config.properties` file to collect these variables? There are too many environment variables for me now.

hmm.. I think it is just a script for testing locally, not for production. I don't prefer to have an additional config file for it. Why not just use `.env` file to maintain them?

grieve54706:
Oh! cool. The `.env` is fine to me. I didn't catch that you meant that before.

grieve54706 (approved):

goldmedal:
Thanks @grieve54706

PR #870, opened 2024-10-31
feat(session-context): cache session context

grieve54706:


goldmedal (approved):
it makes sense to me 👍

PR #871, opened 2024-10-31
chore(function-list): fill functions for each data sources

grieve54706:


grieve54706:
The error from the CI is due to a cached Rust core.

ibis-server/resources/function_list/bigquery.csv
```diff
@@ -0,0 +1,101 @@
...
+scalar,LTRIM,STRING,"Removes leading whitespace."
+scalar,RTRIM,STRING,"Removes trailing whitespace."
+scalar,LENGTH,INT64,"Returns the length of a string."
+scalar,REGEXP_CONTAINS,BOOL,"Returns TRUE if the string contains a match for the regex."
+scalar,REGEXP_EXTRACT,STRING,"Extracts the first match of the regex from the string."
+scalar,REGEXP_REPLACE,STRING,"Replaces all matches of the regex with a replacement string."
+scalar,SUBSTR,STRING,"Returns a substring."
+scalar,CAST,ANY,"Converts a value to a different data type."
+scalar,SAFE_CAST,ANY,"Converts a value to a different data type, returning NULL on error."
+scalar,CURRENT_TIMESTAMP,TIMESTAMP,"Returns the current timestamp."
```
goldmedal, line 76:
```suggestion
```
`CURRENT_TIMESTAMP` should be banned in our case.

grieve54706: It should be banned in the AI service. Not here. It's an exception.

ibis-server/tests/routers/v3/connector/bigquery/test_functions.py
```diff
@@ -0,0 +1,109 @@
...
+        response = client.get(url=f"{base_url}/functions")
+        assert response.status_code == 200
+        result = response.json()
+        assert len(result) == 258
+
+        config.set_remote_function_list_path(function_list_path)
+        response = client.get(url=f"{base_url}/functions")
+        assert response.status_code == 200
+        result = response.json()
+        assert len(result) > 258
```
goldmedal, line 60: I think we should know how many functions are supported exactly.

ibis-server/resources/function_list/bigquery.csv
goldmedal: No, Wren AI service will get the list as their white list. So, we can't add this function to the list.

grieve54706: Should this function be removed only in BigQuery or all data source?

goldmedal: I think we should remove all of them before we make sure they work well.

grieve54706:
I will submit another PR to test all functions.

goldmedal (approved):
Thanks @grieve54706  it's very nice. 👍

Issue #671, opened 2024-07-16
TrinoContainer may encounter empty nodes


grieve54706:
Trino version 450.

After starting the TrinoContainer, we may encounter this error when sending a SELECT statement.
```
2024-07-15 17:15:28 2024-07-15T09:15:28.165Z    INFO    main    io.trino.server.Server  ======== SERVER STARTED ========
2024-07-15 17:15:28 2024-07-15T09:15:28.882Z    INFO    dispatcher-query-3      io.trino.event.QueryMonitor     TIMELINE: Query 20240715_091528_00000_8ufrr :: FINISHED :: elapsed 406ms :: planning 121ms :: waiting 36ms :: scheduling 244ms :: running 32ms :: finishing 9ms :: begin 2024-07-15T09:15:28.446Z :: end 2024-07-15T09:15:28.852Z
2024-07-15 17:15:29 2024-07-15T09:15:29.207Z    INFO    dispatcher-query-3      io.trino.event.QueryMonitor     TIMELINE: Query 20240715_091529_00001_8ufrr :: FINISHED :: elapsed 32ms :: planning 0ms :: waiting 1ms :: scheduling 32ms :: running 0ms :: finishing 32ms :: begin 2024-07-15T09:15:29.172Z :: end 2024-07-15T09:15:29.204Z
2024-07-15 17:15:29 2024-07-15T09:15:29.711Z    INFO    dispatcher-query-1      io.trino.event.QueryMonitor     TIMELINE: Query 20240715_091529_00002_8ufrr :: FINISHED :: elapsed 476ms :: planning 291ms :: waiting 142ms :: scheduling 168ms :: running 12ms :: finishing 5ms :: begin 2024-07-15T09:15:29.232Z :: end 2024-07-15T09:15:29.708Z
2024-07-15 17:15:29 2024-07-15T09:15:29.751Z    INFO    dispatcher-query-1      io.trino.event.QueryMonitor     TIMELINE: Query 20240715_091529_00003_8ufrr :: FINISHED :: elapsed 16ms :: planning 0ms :: waiting 0ms :: scheduling 16ms :: running 0ms :: finishing 16ms :: begin 2024-07-15T09:15:29.734Z :: end 2024-07-15T09:15:29.750Z
[...]

PR #879, opened 2024-11-01
ci(ibis): append Trino test to standard tests

grieve54706:
By introducing a delay after the Trino container startup, we resolved the issue https://github.com/Canner/wren-engine/issues/671.
So we can reintegrate Trino tests into the standard tests.


goldmedal (approved):
👍

PR #880, opened 2024-11-01
ci(ibis): update core to avoid caching

grieve54706:
Sometimes, we update the Rust code, but other existing PRs do not update the Rust core. This may be because of the cache.


goldmedal (approved):
it makes sense to me.

PR #884, opened 2024-11-04
ci(image): split ibis building to two jobs

grieve54706:
We found if the job only builds one platform, it takes 16 minutes. So we can split the two platforms building into two jobs and let them be parallel to reduce the building time.

And also upgrade the `setup-java` action version to avoid warnings.
https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/


goldmedal (approved):
Thanks @grieve54706 LGTM

PR #886, opened 2024-11-04
ci: use macos-arm64 runner

grieve54706:
We found that building a Docker image is slow on the Arm64 platform. May use an Arm64 OS image to build the image will be faster.

GitHub runner image list
https://github.com/actions/runner-images?tab=readme-ov-file#available-images


grieve54706:
MacOS does not include docker.

PR #889, opened 2024-11-05
fix(mssql): escape special characters for odbc

grieve54706:
We found the error 
```
'IM002', '[IM002] [unixODBC][Driver Manager]Data source name not found and no default driver specified (0) (SQLDriverConnect)'
```
Because the password of mssql includes special characters like `{R;3G1/8Al2AniRye` that start with `{` or include `;`.
It should be covered by `{` and `}` and replace `}` with`}}`.

Reference:
https://github.com/mkleehammer/pyodbc/wiki/Connecting-to-databases
https://stackoverflow.com/questions/78531086/pyodbc-connection-string-correctly-escaping-password-with-special-characters/78532507#78532507

### DIFF
--- a/ibis-server/app/model/data_source.py
+++ b/ibis-server/app/model/data_source.py
@@ -166,7 +166,4 @@ class DataSourceExtension(Enum):
 
     @staticmethod
     def _escape_special_characters_for_odbc(value: str) -> str:
-        if ";" in value or value.startswith("{"):
-            return "{" + value.replace("}", "}}") + "}"
-        else:
-            return value
+        return "{" + value.replace("}", "}}") + "}"
--- a/ibis-server/app/model/data_source.py
+++ b/ibis-server/app/model/data_source.py
@@ -108,14 +108,16 @@ class DataSourceExtension(Enum):
             password=info.password.get_secret_value(),
         )
 
-    @staticmethod
-    def get_mssql_connection(info: MSSqlConnectionInfo) -> BaseBackend:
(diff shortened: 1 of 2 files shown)


ibis-server/app/model/data_source.py
```diff
@@ -161,3 +163,10 @@ def get_trino_connection(info: TrinoConnectionInfo) -> BaseBackend:
             user=(info.user and info.user.get_secret_value()),
             password=(info.password and info.password.get_secret_value()),
         )
+
+    @staticmethod
+    def _escape_special_characters_for_odbc(value: str) -> str:
+        if ";" in value or value.startswith("{"):
+            return "{" + value.replace("}", "}}") + "}"
+        else:
+            return value
```
goldmedal, line 172: Could we always add brackets? Maybe we don't need to check if it contains any illegal character? 🤔

grieve54706: It seems to be okay.

grieve54706:
The connection string test should wait until ibis handles special characters by https://github.com/ibis-project/ibis/pull/10437.

goldmedal (approved):
Thanks @grieve54706 LGTM

goldmedal:
> The connection string test should wait until ibis handles special characters by [ibis-project/ibis#10437](https://github.com/ibis-project/ibis/pull/10437).

filed https://github.com/Canner/wren-engine/issues/890 for it

PR #892, opened 2024-11-06
test(mssql): pull out test case about special characters

grieve54706:
Pull out test case to avoid affect other cases


goldmedal (approved):
Thanks @grieve54706 It makes sense to me.

PR #893, opened 2024-11-07
ci(image): adjust build flow

grieve54706:
We found the `amd64` image(first done) will be covered by the `arm64` image(later).
~~We want to keep them by different runners for different arch os runners in the future.~~

Adjust by following this page from Docker official
https://docs.docker.com/build/ci/github-actions/multi-platform

We start to use the GitHub beta arm runner.
https://github.blog/changelog/2024-06-24-github-actions-ubuntu-24-04-image-now-available-for-arm64-runners/

And I also merge `build-dev-image.yml` into `build-image.yml`. They are so similar.


goldmedal (approved):
Thanks @grieve54706 it's really nice. 👍

PR #901, opened 2024-11-12
feat(core): introduce Timestamp Simplify rule to handle timezone

goldmedal:
# Description
We will simplify the timestamp literal when planning SQL to support the complex timezone issues in the different data sources. Something like
- MSSQL only accepts the time zone offset.
- Clikchouse only accepts the name of the time zone.
- BigQuery has timestamp (impacted by timezone) and Datetime (non-impacted by timezone).

To handle them, Wren Core will convert all of the time zone literals to UTC, which is how other databases process the timestamp with the time zone literal.

Powered by DataFusion, both the offset and the timezone name are allowed. It can also handle the DST issue.

## How the SQL transformed
When you input a SQL like
```
select timestamp '2011-01-01 18:00:00 Asia/Taipei'
```

The planned SQL will be
```
SELECT CAST('2011-01-01 10:00:00' AS TIMESTAMP)
```

Then, the dialect will be if the dialect is BigQuery.
```
SELECT CAST('2011-01-01 10:00:00' AS DATETIME)
```

### DIFF
--- a/ibis-server/tests/routers/v3/connector/bigquery/test_query.py
+++ b/ibis-server/tests/routers/v3/connector/bigquery/test_query.py
@@ -5,7 +5,7 @@ import pytest
 from fastapi.testclient import TestClient
 
 from app.main import app
-from tests.routers.v3.connector.postgres.conftest import base_url
+from tests.routers.v3.connector.bigquery.conftest import base_url
 
 manifest = {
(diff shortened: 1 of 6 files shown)


grieve54706 (approved):
LGTM. Thanks for the helpful feature.

PR #903, opened 2024-11-13
ci(image): fix missing sha tag

grieve54706:
We found the sha tag missing. Because we cover the text by `>` when writing to the `tags.txt`, it should be `>>` to append it.

I also found an entry `type=schedule` that can delegate `nightly`. 
https://github.com/docker/metadata-action?tab=readme-ov-file#typeschedule


goldmedal (approved):
Thanks @grieve54706  for fixing it.

PR #905, opened 2024-11-13
ci: test bq credential

grieve54706:


grieve54706:
The credential is good.

PR #907, opened 2024-11-13
fix(core): disable pushdown filter to keep the casting for timestamp column

goldmedal:
# Description
Sometimes, we will cast a timestamp column to a timestamp type because we must handle the `DATETIME` and `TIMESTAMP` for the BigQuery. However, `push_down_filter` will remove the casting because it can't differentiate `DATETIME` and `TIMESTAMP` in the DataFusion layer.

see the unit test for the details.
If applying `push_down_filter`, the result will be as below. However, it will fail if `出道時間` is `DATETIME` in BigQuery
```sql
SELECT 
  count(*) 
FROM 
  (
    SELECT 
      artist.cast_timestamp 
    FROM 
      (
        SELECT 
          CAST(
            artist."出道時間" AS TIMESTAMP WITH TIME ZONE
          ) AS cast_timestamp 
        FROM 
          artist 
        WHERE 
          artist."出道時間" > CAST(
            '2011-01-01 21:00:00' AS TIMESTAMP
          )
      ) AS artist
  ) AS artist
```


grieve54706 (approved):
LGTM

PR #908, opened 2024-11-14
feat(core): register timestamptz type to DataFusion timestamp with time zone type

goldmedal:
# Description
Now we can distinguish `TIMESTAMP ` and `TIMESTAMP WITH TIME ZONE` in the DataFusion layer. We also know the sqlglot will transpile SQL for Bigquery
- `TIMESTAMP` -> `DATETIME` ( without timezone)
- `TIMESTAMP WITH TIME ZONE` -> `TIMESTAMP` ( with timezone)

We can apply DataFusion's type coercion rule for them. If needed, DataFusion will cast the timestamp side to the timestamp with the time zone.

## Additional change
- Enhance the validation script. Now we can set the log level for the detailed debug information.
```
LOGURU_LEVEL=DEBUG python tools/mdl_validation.py mdl.json function.csv
```
- add `boolean` alias for bool type
- add required function `current_datetime` for BigQuery


wren-core/core/src/logical_plan/utils.rs
```diff
@@ -79,7 +81,6 @@ pub fn map_data_type(data_type: &str) -> DataType {
         // BigQuery Compatible Types
         "bignumeric" => DataType::Decimal128(38, 10), // set the default precision and scale
         "bytes" => DataType::Binary,
-        "datetime" => DataType::Timestamp(TimeUnit::Nanosecond, None), // chose the smallest time unit
```
grieve54706, line 82: I'm curious, why remove datetime?

goldmedal: Just merge with `timestamp` at L69

grieve54706: I saw that, thanks.

grieve54706 (approved):
LGTM. Thanks @goldmedal.

goldmedal:
Thanks @grieve54706  for reviewing

PR #925, opened 2024-11-19
feat(core): support to define param_names and param_types in the list file

goldmedal:
# Description
Because the CSV doesn't support the native array format, use a comma-separated string to present the list.
## Sample
```
function_type,name,return_type,param_names,param_types,description
scalar,add_two,int,"f1,f2","int,int","Adds two numbers together."
```


grieve54706 (approved):
LGTM. Thanks @goldmedal.

PR #922, opened 2024-11-19
test: create a framework for testing functions

grieve54706:
Test all functions in CSV could be executed.

PR #927, opened 2024-11-20
test: create a framework for testing functions

grieve54706:
Add test framework for testing all functions in CSV
- #922
- #928
- #947 
- #950
- #952
- #954

It's a basic framework. Many functions are not supported yet, but these issues should be fixed in the future.
- #948
- #949
- #951
- #953


goldmedal (approved):
Thanks @grieve54706 It looks good to me.

PR #933, opened 2024-11-20
feat(engine): handle target.*

grieve54706:
Handle the sql with dot all like `SELECT table.* FROM table` and `SELECT t.* FROM table AS t`.


goldmedal:
Thanks @grieve54706.  Could you add some test for the model with calculated fields and the SQL with a join statement?
```
select a.* from a join b on a.c1 = b.c1
```

ibis-server/tests/routers/v2/connector/test_postgres.py
```diff
@@ -148,6 +148,23 @@ def test_query_with_connection_url(manifest_str, postgres: PostgresContainer):
...
+        connection_info = _to_connection_info(postgres)
+        response = client.post(
+            url=f"{base_url}/query",
+            params={"limit": 1},
+            json={
+                "connectionInfo": connection_info,
+                "manifestStr": manifest_str,
+                "sql": 'SELECT "o".* FROM "Orders" AS "o"',
+            },
+        )
```
goldmedal, line 161: Could you add the end-to-end SQL test for the calculated field and join syntax? Maybe we should add tested in `TestDynamicFields`, too. We need to make sure the model-generating behavior is fine in `WrenSqlRewrite`.

goldmedal (approved):
Thanks @grieve54706 it looks nice to me.

PR #943, opened 2024-11-25
fix(core): make aggregate functions and window functions support zero argument

### LINKED ISSUE #924: Wren core transform SQL failed `Error during planning: cume_dist does not support zero arguments`
```sql
SELECT
    cume_dist() OVER (ORDER BY id) AS cume_dist 
FROM (
    SELECT 1 AS id, 'A' AS category UNION ALL
    SELECT 2 AS id, 'B' AS category UNION ALL
    SELECT 3 AS id, 'A' AS category UNION ALL
    SELECT 4 AS id, 'B' AS category UNION ALL
    SELECT 5 AS id, 'A' AS category
) AS t
```
```
DataFusion error: Error during planning: Error during planning: cume_dist does not support zero arguments. No function matches the given name and argument types 'cume_dist()'. You might need to add explicit type casts.
	Candidate functions:
	cume_dist(Any, .., Any)
```

Also other window functions too.
`dense_rank`, `percent_rank`, `rank`, `row_number`.

BigQuery function reference
https://cloud.google.com/bigquery/docs/reference/standard-sql/numbering_functions#cume_dist

grieve54706:
Resolve #924

### DIFF
--- a/wren-core/core/src/mdl/function.rs
+++ b/wren-core/core/src/mdl/function.rs
@@ -117,7 +117,13 @@ impl ByPassAggregateUDF {
         Self {
             name: name.to_string(),
             return_type,
-            signature: Signature::variadic_any(Volatility::Immutable),
+            signature: Signature::one_of(
+                vec![
+                    TypeSignature::VariadicAny,
+                    TypeSignature::Uniform(0, vec![]),
+                ],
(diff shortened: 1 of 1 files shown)


goldmedal (approved):
Thanks @grieve54706 LGTM

Issue #953, opened 2024-11-26
Add functions for Trino


grieve54706:
We added a test case for all functions. But don't have time to fix all. So we move out some functions that need to be revised and submit this issue to add them back.
```csv
function_type,name,return_type,param_names,param_types,description
aggregate,approx_percentile,same as input,,array<any>,Approximates percentile
aggregate,approximate_distinct,bigint,,any,Approximates count of distinct values
scalar,current_date,date,,none,Returns current date
scalar,date_add,date,,varchar,interval,Adds interval to date
scalar,date_sub,date,,varchar,interval,Subtracts interval from date
scalar,date_trunc,timestamp,,varchar,timestamp,Truncates timestamp to specified precision
scalar,element_at,any,,array,integer,Returns element at specified position in array
scalar,filter,array,,array,function,Filters array using lambda
scalar,format_datetime,varchar,,timestamp,varchar,Formats datetime according to format string
scalar,from_hex,varbinary,,varchar,Converts hex string to binary
scalar,json_format,varchar,,json,Pretty prints JSON
scalar,json_parse,json,,varchar,Parses string as JSON
scalar,length,bigint,,varchar,Returns length of string
scalar,md5,varchar,,varchar,Computes MD5 hash
scalar,parse_datetime,timestamp,,varchar,varchar,Parses string to datetime using format
scalar,reduce,any,,array,function,function,any,Reduces array to single value using lambda
scalar,sha256,varchar,,varchar,Computes SHA256 hash
scalar,to_hex,varchar,,bigint,Converts number to hex string
scalar,transform,array,,array,function,Applies lambda to each element
scalar,trim,varchar,,varchar,Removes leading and trailing whitespace
scalar,uuid,varchar,,none,Generates random UUID
scalar,xxhash64,bigint,,varchar,Computes xxHash64 hash
scalar,zip_with,array,,array,array,function,Combines two arrays using lambda
window,cume_dist,double,,none,Returns cumulative distribution
window,dense_rank,bigint,,none,Returns rank without gaps
window,first_value,any,,any,integer,Returns first value in window
window,lag,any,,any,integer,Returns value from previous row
window,last_value,any,,any,integer,Returns last value in window
window,lead,any,,any,integer,Returns value from following row
window,nth_value,any,,any,integer,Returns nth value in window
window,ntile,bigint,,integer,Divides rows into buckets
window,percent_rank,double,,none,Returns percent rank of row
window,rank,bigint,,none,Returns rank with gaps
[...]

PR #954, opened 2024-11-26
test(trino): test all functions

grieve54706:
We added a test case and moved out some functions that need to be revised.

https://github.com/Canner/wren-engine/issues/953


ibis-server/resources/function_list/trino.csv
```diff
@@ -1,86 +1,30 @@
...
-window,nth_value,any,"Returns value at specified row"
-window,ntile,bigint,"Divides rows into buckets"
-window,percent_rank,double,"Returns percent rank of row"
-window,rank,bigint,"Returns rank with gaps"
-window,row_number,bigint,"Returns sequential row number"
+function_type,name,return_type,param_names,param_types,description
+aggregate,array_agg,array,,any,Aggregates values into array
+aggregate,avg,double,,numeric,Returns average of values
+aggregate,count,bigint,,any,Counts number of rows
+aggregate,max,same as input,,any,Returns maximum value
```
goldmedal, line 5:
```suggestion
aggregate,max,same as input,,any,Returns maximum value
```
Similar issue. I think we can remove them.

goldmedal (approved):
Thanks @grieve54706 there is a conflict. I think you can free feel to merge it after fix the conlfict.

PR #956, opened 2024-11-27
feat(core): upgrade DataFusion to support the array of struct syntax

goldmedal:
# Description
- To support the syntax of the array of structs like `a[1].b.c`, I enhanced the `sqlparser-rs` syntax
    - https://github.com/apache/datafusion-sqlparser-rs/pull/1541
- Implement the planner behavior on the DataFusion side
    - https://github.com/goldmedal/datafusion/pull/1
 
Before they are merged back into the DataFusion, Wren core will be based on the specific commit to support the syntax.


grieve54706 (approved):
LGTM

PR #959, opened 2024-11-29
feat(wren-core-py): implement ManifestExtractor

grieve54706:
## Description
We have found that the Java server often times out. We suspect that this is due to processing the analysis of the fat manifest.
So we implement `ManifestExtractor` to extract the used models from the manifest, thereby reducing the Java server's analysis load.
We design a two-step process. The first step is `resolve_used_table_names`, which takes an SQL query and returns the names of the used tables. The second step is `extract_by`, which takes the used table names and returns a new manifest only with used models.

## Example
```python
import wren_core

extractor = wren_core.ManifestExtractor(manifest_str)
tables = extractor.resolve_used_table_names(sql)
manifest = extractor.extract_by(tables)
simple_manifest_str = to_json_base64(manifest)
```


wren-core-py/src/extractor.rs
```diff
@@ -0,0 +1,306 @@
...
+}
+
+fn extract_relationships(
+    mdl: &Arc<WrenMDL>,
+    used_datasets: &[String],
+) -> Vec<Arc<Relationship>> {
+    let mut used_set: HashSet<String> = used_datasets.iter().cloned().collect();
+    let mut stack: Vec<String> = used_datasets.to_vec();
+    while let Some(dataset_name) = stack.pop() {
+        if let Some(relationship) = mdl.get_relationship(&dataset_name) {
```
goldmedal, line 147: I think it's wrong. `get_relationship` is used to look for a relationship based on the name of the relationship. It's not related to the name of the dataset.
You must search the dataset name in the relationship models to ensure if the relationship is required.

goldmedal: By the way, maybe we should add the tests for the relationship.

wren-core-py/src/extractor.rs
```diff
@@ -0,0 +1,306 @@
...
+mod tests {
+    use crate::extractor::PyManifestExtractor;
+    use base64::prelude::BASE64_STANDARD;
+    use base64::Engine;
+    use rstest::{fixture, rstest};
+    use std::iter::Iterator;
+
+    #[fixture]
+    pub fn mdl_base64() -> String {
+        let mdl_json = r#"
```
goldmedal, line 172: I prefer to use `ManifestBuilder` to build the case and then serialize it to the JSON if required. It's easier to maintain.

wren-core-py/Cargo.toml
```diff
@@ -9,19 +9,23 @@ name = "wren_core_py"
...
+[dev-dependencies]
+rstest = "0.23.0"
+
 [build-dependencies]
-pyo3-build-config = "0.21.2"
+pyo3-build-config = "0.23.2"
+
+[features]
+extension-module = ["pyo3/extension-module"]
+default = ["extension-module"]
```
goldmedal, line 31: I'm not sure but why do we need this feature flag? 🤔  I tried to remove them and everything is fine.

goldmedal (commented):
Thanks, @grieve54706. Most of them are good, but I think the relationship extraction is wrong. I also left some suggestions.

wren-core-py/src/extractor.rs
```diff
@@ -0,0 +1,306 @@
...
+                .filter(|t| {
+                    t.catalog().map_or(true, |catalog| catalog == mdl.catalog())
+                        && t.schema().map_or(true, |schema| schema == mdl.schema())
+                })
+                .map(|t| t.table().to_string())
+                .collect()
+        })
+}
+
+pub fn extract_manifest(
```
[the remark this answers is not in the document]
grieve54706: It will be the same as `PyManifestExtractor.extract_by`. Still, add the same doc?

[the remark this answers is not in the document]
goldmedal: Sure. We can also add a reference to the related methods.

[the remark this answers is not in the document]
grieve54706: It should not be public. I'll remove the visibility modifier.

wren-core-py/src/extractor.rs
grieve54706: I will fix it. Thanks.

wren-core-py/src/extractor.rs
grieve54706: Sure.

wren-core-py/Cargo.toml
grieve54706: Follow this [FAQ](https://pyo3.rs/v0.22.5/faq#i-cant-run-cargo-test-or-i-cant-build-in-a-cargo-workspace-im-having-linker-issues-like-symbol-not-found-or-undefined-reference-to-_pyexc_systemerror)
If we add these configurations, we can test Rust without compiling to Python.

wren-core-py/src/extractor.rs
```diff
@@ -0,0 +1,326 @@
...
+                .columns
+                .iter()
+                .filter_map(|col| {
+                    col.relationship
+                        .as_ref()
+                        .and_then(|rel_name| mdl.get_relationship(rel_name))
+                })
+                .flat_map(|rel| rel.models.clone())
+                .filter(|related| used_set.insert(related.clone()))
+                .for_each(|related| stack.push(related));
```
goldmedal, line 101: We can change`used_set` to a HashMap
```rust
    let mut used_set: HashMap<String, usize> = used_datasets.iter().map(|s| (s.clone(), 0)).collect();
```
Then, we can use `Entry` to avoid cloning `related` if it has existed.
```suggestion
                .for_each(|related| {
                    if let Entry::Vacant(vacant) = used_set.entry(related) {
                        let key = vacant.key().clone();
                        vacant.insert(0);
                        stack.push(key);
                    }
                });
```

goldmedal (approved):
Thank @grieve54706.  Overall looks good to me. I left some improvement suggestions.

wren-core-py/src/extractor.rs
grieve54706: If we care about clone stack, we can use `HashSet:: contains`.
```suggestion
                .for_each(|related| {
                    if !used_set.contains(&related) {
                        used_set.insert(related.clone());
                        stack.push(related);
                    }
                });
```

Issue #930, opened 2024-11-20
Wren core over rewrite `INTERVAL`


grieve54706:
Source
```sql
SELECT date_add(cast('2024-01-01' as date), INTERVAL 1 day)
```
Planned
```sql
SELECT date_add(CAST('2024-01-01' AS DATE), INTERVAL '1 0:0:0.000' DAY TO SECOND)
```
Sqlglot can not receive it
```
sqlglot.errors.ParseError: Required keyword: 'this' missing for <class 'sqlglot.expressions.DateAdd'>. Line 1, Col: 81.
  SELECT date_add(CAST('2024-01-01' AS DATE), INTERVAL '1 0:0:0.000' DAY TO SECOND)
```

PR #961, opened 2024-12-02
feat(core): use MySQL interval style for output format

goldmedal:
# Description
Part of #930 
Refer to [the interval style of DataFusion](https://github.com/apache/datafusion/blob/172e5573270ae4a3d152d5bf15c5c0008595090e/datafusion/sql/src/unparser/dialect.rs#L171), `MySQL` style is more common for various data sources.


grieve54706 (approved):
LGTM. Thanks.

PR #962, opened 2024-12-02
feat(ibis): use the default input dialect

### LINKED ISSUE #930: Wren core over rewrite `INTERVAL`

goldmedal:
# Description
Close #930
The root cause of #930 is the `trino` dialect has a specific limit for the [date_add](https://trino.io/docs/current/functions/datetime.html#date_add) function. We can't use it as a normal scalar function. We should follow the spec of trino doc.
```python
import sqlglot
sql = "SELECT date_add('second', 1, CAST('2024-01-01' AS DATE))"
result = sqlglot.transpile(sql, read="trino", write="bigquery")[0]
print(result)
# SELECT DATE_ADD(CAST('2024-01-01' AS DATE), INTERVAL 1 SECOND)
```

However, we want a common SQL base for some general purpose. I think we can change the input dialect to the default dialect which has fewer limitations for the function usage. 
```python
import sqlglot
sql = "SELECT date_add(CAST('2024-01-01' AS DATE), INTERVAL 1 DAY)"
result = sqlglot.transpile(sql, write="bigquery")[0]
print(result)
# SELECT DATE_ADD(CAST('2024-01-01' AS DATE), INTERVAL (INTERVAL '1' DAY) DAY)
```

However, the output for the BigQuery is still wrong. `INTERVAL (INTERVAL '1' DAY) DAY)` isn't a valid SQL for BigQuery. It seems SQLGlot has some specific definition for the `date_add` function. The SQL should be
```python
import sqlglot
sql = "SELECT date_add(CAST('2024-01-01' AS DATE), 1)"
result = sqlglot.transpile(sql, write="bigquery")[0]
print(result)
# SELECT DATE_ADD(CAST('2024-01-01' AS DATE), INTERVAL 1 DAY)
```
It's different from [the original definition of BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/date_functions#date_add)

# Additional Changed
- Modified the `date_add` definition for bigquery
- Disable `instr` for MySQL
    - Because `instr` is an alias of `strpos` in DataFusion, the input `instr` function will be transformed to `strpos`, then the sqlglot can't transpile `strpos`to `instr` when using the default dialect to `mysql` dialect`. We need to file another issue for it.


ibis-server/app/mdl/rewriter.py
```diff
@@ -44,7 +44,7 @@ def rewrite(self, sql: str) -> str:
 
     def _transpile(self, planned_sql: str) -> str:
         write = self._get_write_dialect(self.data_source)
-        return sqlglot.transpile(planned_sql, read="trino", write=write)[0]
+        return sqlglot.transpile(planned_sql, write=write)[0]
```
grieve54706, line 47: This method is used for v2 and v3. For v2, the read dialect is `trino`. How about splitting this method for each class?

goldmedal: hmm..  No, maybe you're right. We're better to split v2 and v3 because the trino dialect also handles the functions converting between trino and others data source ( Currently, the v2 function white list used by AI service is based on the trion function) 🤔

goldmedal:
The BigQuery CI has issues. I tested BigQuery in my local environment and all pass.
```
$ just test bigquery
poetry run pytest -m 'bigquery'
============================================================================================================================== test session starts ===============================================================================================================================
platform darwin -- Python 3.11.6, pytest-8.3.3, pluggy-1.5.0
rootdir: /ibis-server
configfile: pyproject.toml
plugins: anyio-4.6.2.post1
collected 205 items / 175 deselected / 30 selected                                                                                                                                                                                                                               

tests/routers/v2/connector/test_bigquery.py .................                                                                                                                                                                                                              [ 56%]
tests/routers/v3/connector/bigquery/test_functions.py ....                                                                                                                                                                                                                 [ 70%]
tests/routers/v3/connector/bigquery/test_query.py .........                                                                                                                                                                                                                [100%]

================================================================================================================= 30 passed, 175 deselected in 283.35s (0:04:43) =================================================================================================================
```

grieve54706 (approved):
LGTM

PR #972, opened 2024-12-06
fix(core): fix the StatementAnalyzer for UNNEST and LATERAL query

goldmedal:
# Description
We're missing to analyze the expression in the `UNNEST` and `LATERAL`. It may cause the wrong model generation. The following SQL would be failed:
- `select u.a1 from nested_table n, unnest(n.a1) u(a1)`
- `select l.a1 from nested_table n cross join lateral (select n.a1[1]) l(a1)`


grieve54706 (approved):
LGTM. Thanks @goldmedal.

PR #974, opened 2024-12-09
feat(java): warm up java engine

grieve54706:
We noticed that the first request to the Java engine is often timed out.
The Ibis server calls the Java engine timeout `5s`.
So we warm up the Java engine when it starts.
The warm-up time: ~5s. The warm-up case can not be too simple like `SELECT 1`. It will be ineffective.

### Benchmark
Docker resource limit
```yaml
java-engine:
  deploy:
    resources:
      limits:
        cpus: '0.5'
        memory: 2G
```
API: `POST {{ibis-server}}/v2/connector/dry-plan`
- Before: timeout(over 5s)
- After: 1.52s, 1.65s, 1.79s, 1.92s

### Additional information
[Java warm up](https://www.baeldung.com/java-jvm-warmup)
[python httpx timeouts](https://www.python-httpx.org/advanced/timeouts)
[docker compose resource](https://docs.docker.com/reference/compose-file/deploy/#resources)

### DIFF
--- a/wren-main/src/main/java/io/wren/main/PreviewService.java
+++ b/wren-main/src/main/java/io/wren/main/PreviewService.java
@@ -28,6 +28,7 @@
 
 package io.wren.main;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Streams;
 import com.google.inject.Inject;
 import io.airlift.log.Logger;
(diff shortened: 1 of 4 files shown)


goldmedal (approved):
Thanks @grieve54706  let's try it 🚀

PR #979, opened 2024-12-10
feat(ibis): use `httpx.AsyncClient` instead of httpx.request

grieve54706:
Use `httpx.AsyncClient` to enhance performance.
- Use `httpx.Client` instead of `httpx.request` to keep the client for reducing handshake for every request.
- Use `httpx.AsyncClient` for better performance.
- Make test class with AsyncClient

Additional information
https://fastapi.tiangolo.com/advanced/async-tests/
https://fastapi.tiangolo.com/async/
https://anyio.readthedocs.io/en/stable/testing.html

### DIFF
--- a/ibis-server/app/mdl/rewriter.py
+++ b/ibis-server/app/mdl/rewriter.py
@@ -1,3 +1,4 @@
+import asyncio
 import importlib
 
 import httpx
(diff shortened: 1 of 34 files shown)


ibis-server/app/mdl/rewriter.py
```diff
@@ -95,14 +95,14 @@ def __init__(self, manifest_str: str, function_path: str):
...
-    def rewrite(self, sql: str) -> str:
+    async def rewrite(self, sql: str) -> str:
         try:
             extractor = get_manifest_extractor(self.manifest_str)
             tables = extractor.resolve_used_table_names(sql)
             manifest = extractor.extract_by(tables)
             manifest_str = to_json_base64(manifest)
             session_context = get_session_context(manifest_str, self.function_path)
-            return session_context.transform_sql(sql)
+            return await to_thread.run_sync(session_context.transform_sql, sql)
```
goldmedal, line 105: It's potentially a nice improvement. `transform_sql` is an async Rust method, but we wrap it into a sync method. I found pyo3 has [async interface](https://pyo3.rs/v0.13.2/ecosystem/async-await). Maybe we can change to it in the future.

goldmedal (approved):
Thanks @grieve54706, Overall looks good to me. Nice work.

PR #980, opened 2024-12-12
feat(core): add `dataSource` field to MDL

### LINKED ISSUE #970: Add `dataSource` field in the MDL
Sometimes, we need to change our behavior according to the kind of data source. It's good to add a new field for it.

- Add a new field in `Manifest`
```
pub data_source: Option<String>
```
- Implement the mapping function for the supported data source.
```rust
pub fn data_source_from_str(name: impl AsRef<str>) -> Option<String> {
    match name.to_lowercase().as_str() {
        "bigquery" => Some("bigquery"),
        ...
         _ => None,
    }
}
```
The return value couldn't be `String`. It may be an enum or something. Just write `String` temporarily.

goldmedal:
close #970

### DIFF
--- a/wren-core/core/src/mdl/manifest.rs
+++ b/wren-core/core/src/mdl/manifest.rs
@@ -18,6 +18,50 @@ pub struct Manifest {
     pub metrics: Vec<Arc<Metric>>,
     #[serde(default)]
     pub views: Vec<Arc<View>>,
+    #[serde(default)]
+    pub data_source: Option<DataSource>,
+}
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Hash, Clone, Copy)]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
+pub enum DataSource {
+    #[serde(alias = "bigquery")]
+    BigQuery,
+    #[serde(alias = "clickhouse")]
+    Clickhouse,
+    #[serde(alias = "canner")]
+    Canner,
+    #[serde(alias = "trino")]
+    Trino,
+    #[serde(alias = "mssql")]
+    MSSQL,
+    #[serde(alias = "mysql")]
+    MySQL,
+    #[serde(alias = "postgres")]
+    Postgres,
+    #[serde(alias = "snowflake")]
+    Snowflake,
+    #[serde(alias = "datafusion")]
+    Datafusion,
+    #[serde(alias = "duckdb")]
+    DuckDB,
+}
+
+impl Display for DataSource {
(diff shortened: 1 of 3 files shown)


wren-core/core/src/mdl/manifest.rs
```diff
@@ -18,6 +18,50 @@ pub struct Manifest {
...
+            DataSource::BigQuery => write!(f, "BIGQUERY"),
+            DataSource::Clickhouse => write!(f, "CLICKHOUSE"),
+            DataSource::Canner => write!(f, "CANNER"),
+            DataSource::Trino => write!(f, "TRINO"),
+            DataSource::MSSQL => write!(f, "MSSQL"),
+            DataSource::MySQL => write!(f, "MYSQL"),
+            DataSource::Postgres => write!(f, "POSTGRES"),
+            DataSource::Snowflake => write!(f, "SNOWFLAKE"),
+            DataSource::Datafusion => write!(f, "DATAFUSION"),
+            DataSource::DuckDB => write!(f, "DUCKDB"),
```
grieve54706, line 62: Curious, why make them uppercase?

grieve54706 (approved):

grieve54706:
Could add a test case to assert this feature?

wren-core/core/src/mdl/manifest.rs
goldmedal: No, specific reason. Just use the same style as the relationship type for the enum value.

goldmedal:
> Could add a test case to assert this feature?

I added tests in `test_manifest_roundtrip` and the JSON file of the mdl to test if it can be serialized and deserialized well.

wren-core/core/src/mdl/manifest.rs
grieve54706: The ibis-server and web use lowercase. Does this change will not affect the JSON format?

goldmedal: I think it won't affect it. We also accept the lowercase value.

grieve54706: When the ibis-server calls `ManifestExtractor` and then gets the object with an uppercase data source name, can it map to the `DataSource` enum of ibis-server?

goldmedal: It depends on how to implement it. By the way, why do we need to map to the enum of ibis-server?

goldmedal: It's fine for me to change them to lowercase if you think it's better. It doesn't matter actually. 🤔

goldmedal: Thanks

PR #984, opened 2024-12-13
feat(core): support querying window function

### LINKED ISSUE #973: Support querying `Window` function
If a query using the window function to access a model, the field in the window function can't be recognized correctly.
```
SELECT sum(c_custkey) over (order by c_name) FROM test.test.customer limit 1
```

Given a MDL and query the above SQL:
```json
{
  "catalog": "test",
  "schema": "test",
  "models": [
    {
      "name": "customer",
      "tableReference": {
        "catalog": "",
        "schema": "",
        "table": "customer"
      },
      "columns": [
        {
          "name": "c_custkey",
          "type": "integer"
        },
        {
          "name": "c_name",
          "type": "varchar"
        }
      ]
    }
  ]
}
```

We will get the error message
```
Error: ModelAnalyzeRule
caused by
Schema error: No field named test.test.customer.c_custkey.
```

The root cause is that we are missing to handle `LogicalPlan::Window` at 

https://github.com/Canner/wren-engine/blob/04f70154e1f10c65ea908702e86d1d3b1b611931/wren-core/core/src/logical_plan/analyze/model_anlayze.rs#L116
We must handle the `LogicalPlan::Window` plan and collect the required columns.
```
LogicalPlan::Window(window) => {
    todo()!
}
```

Then, add tests in `mdl/mod.rs`
https://github.com/Canner/wren-engine/blob/6fd070dd48843795229b47cc7549bbe7da27fb21/wren-core/core/src/mdl/mod.rs#L449

and
`sqllogictests/test_files/model.lst`

https://github.com/Canner/wren-engine/blob/04f70154e1f10c65ea908702e86d1d3b1b611931/wren-core/sqllogictest/test_files/model.slt#L33

grieve54706:
Resolve #973

### DIFF
--- a/wren-core/core/src/logical_plan/analyze/model_anlayze.rs
+++ b/wren-core/core/src/logical_plan/analyze/model_anlayze.rs
@@ -223,6 +223,24 @@ impl ModelAnalyzeRule {
                 scope_mut.add_visited_table(subquery_alias.alias.clone());
                 Ok(Transformed::no(plan))
             }
+            LogicalPlan::Window(window) => {
+                let mut scope_mut = scope.borrow_mut();
+                window
+                    .window_expr
+                    .iter()
+                    .fold(HashSet::new(), |mut set, expr| {
+                        Expr::add_column_refs(expr, &mut set);
+                        set
+                    })
+                    .into_iter()
+                    .try_for_each(|col| {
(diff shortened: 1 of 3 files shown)


goldmedal (approved):
Thanks @grieve54706 LGTM

PR #986, opened 2024-12-16
chore: wrap http client to init in lifespan

grieve54706:
The client started should be in lifespan gather stoped.

We need to wrap the client into a class. We call it `JavaEngineConnector`.
And make it a singleton.

Also trigger lifespan in the test, we need [asgi-lifespan](https://github.com/florimondmanca/asgi-lifespan).
This library is recommended by FastAPI, as noted in the warning message on the fastapi [async-tests](https://fastapi.tiangolo.com/advanced/async-tests/#in-detail) page.


goldmedal (approved):
Thanks @grieve54706 LGTM

PR #988, opened 2024-12-16
chore: use anyio instead of asyncio

grieve54706:


goldmedal (approved):
Could you explain why did we switch to `anyio`?

grieve54706:
I have another way to avoid using `asyncio.run`.

PR #993, opened 2024-12-18
fix(core): remove unused check to fix union syntax

goldmedal:
# Description
Because the scope analysis isn't fully finished in the Java engine, the check for the union schema is wrong. The Java engine's main purpose is to be a pure SQL planner. We don't need to do too many checks. Just focus on collecting which columns and models are required.
This PR removes the checks for the union's input to ensure the union works well.


grieve54706 (approved):
LGTM

PR #994, opened 2024-12-18
ci(release): auto generate gh release note

grieve54706:
Auto-generate gh release note when stable-releasing

Ref: https://github.com/softprops/action-gh-release


goldmedal (approved):
It looks nice. Thanks.

PR #998, opened 2024-12-19
chore: add class State and use state instead of singleton

grieve54706:
Add class `State` for lifespan and use state of lifespan instead of Singleton and remove unused `get_java_engine_connector` function

### DIFF
--- a/ibis-server/app/main.py
+++ b/ibis-server/app/main.py
@@ -1,5 +1,6 @@
 from collections.abc import AsyncIterator
 from contextlib import asynccontextmanager
+from typing import TypedDict
 from uuid import uuid4
 
 from asgi_correlation_id import CorrelationIdMiddleware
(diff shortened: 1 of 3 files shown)


goldmedal (approved):

ibis-server/app/main.py
```diff
@@ -9,17 +10,21 @@
...
+from app.mdl.java_engine import JavaEngineConnector
 from app.middleware import ProcessTimeMiddleware, RequestLogMiddleware
 from app.model import ConfigModel, CustomHttpError
 from app.routers import v2, v3
 
 get_config().init_logger()
 
 
+class State(TypedDict):
+    java_engine_connector: JavaEngineConnector
```
goldmedal, line 22: It's better to add some comment or explanation for this class.

PR #999, opened 2024-12-19
feat(core): implement the dialect-specific function unparsing

### LINKED ISSUE #946: Incorrectly transform SQL from `trim` to `btrim`
Source
```sql
SELECT trim('test')
```
Transformed
```sql
SELECT btrim('test')
```
The `btrim` is a function of datafusion. But MySQL does not support it.

goldmedal:
# Description
close #946 
The root cause of #946 is `trim` is the alias of `btrim` in the DataFusion. The `btrim` function call plan will be generated when we plan the SQL to the logical plan. That's why we got the `btrim` in the unparsing result. In this case, we need to override the behavior of SQL generation. This PR  supports the `trim` function for MySQL. We can follow this pattern to support the similar cases for other data sources.

# Changed
- introduce `InnerDialect` to customize the function unparsing behavior for different data source.
- Fix the data_source serde
- Add `trim` to the function white list of MySQL


wren-core-py/tests/test_modeling_core.py
```diff
@@ -193,6 +194,7 @@ def test_extract_by(dataset, expected_models):
     extracted_manifest = ManifestExtractor(manifest_str).extract_by(dataset)
     assert len(extracted_manifest.models) == len(expected_models)
     assert [m.name for m in extracted_manifest.models] == expected_models
+    assert extracted_manifest.data_source.__str__() == "DataSource.BigQuery"
```
grieve54706, line 197: It changes the `data_source` type. Could we keep the field as a string when deserding in the core?

goldmedal: > It changes the data_source type
>
What did you mean by this? I think it's still an enum but invoking to_string here.

grieve54706: I mean it should be a string, not an enum. Because the `data_source` in the source manifest is a string.
And I am not sure the enum from the core could equal the `app.model.data_source.DataSource`.

goldmedal: >I mean it should be a string, not an enum. Because the data_source in the source manifest is a string. 
>

`extracted_manifest` is a python class. Its `data_source` is an enum. The `data_source` will never be a string after deserialized from a JSON to the class.

> And I am not sure the enum from the core could equal the app.model.data_source.DataSource
>
Here is just a way to assert the data source to still be `BigQuery` after extraction. It's not related to the JSON serialize and deserialize.

grieve54706 (approved):
LGTM

PR #1005, opened 2024-12-23
feat(extract): handle extraction failed

### LINKED ISSUE #1004: Fallback to the original path if manifest extracting fail
The following SQL is valid for Trino (v2 dialect) but not valid for the DataFusion (v3 dialect)
```sql
SELECT 
  "c"."properties_industry" AS "industry", 
  COUNT(
    "d"."properties_hs_is_closed_won"
  ) FILTER(
    WHERE 
      "d"."properties_hs_is_closed_won" = TRUE
  ) * 1.0 / COUNT(
    "d"."properties_hs_is_closed_won"
  ) AS "success_rate" 
FROM 
  "deals" AS "d" 
  JOIN "deals_companies" AS "dc" ON "d"."id" = "dc"."deal_id" 
  JOIN "companies" AS "c" ON "dc"."company_id" = "c"."id" 
GROUP BY 
  "c"."properties_industry" 
ORDER BY 
  "success_rate" DESC
```
However, this SQL should work for v2 API

I think it's an edge case for Manifest extracting. We can return the original manifest if the extracting fails to ensure the query flow is fine.

grieve54706:
Resolve #1004
In v2, log the error and use the original manifest when it failed
However, in v3, the error is still raised.

### DIFF
--- a/ibis-server/app/mdl/rewriter.py
+++ b/ibis-server/app/mdl/rewriter.py
@@ -45,7 +45,9 @@ class Rewriter:
         return sqlglot.transpile(planned_sql, read=read, write=write)[0]
 
     async def rewrite(self, sql: str) -> str:
-        manifest_str = self._extract_manifest(self.manifest_str, sql)
+        manifest_str = (
+            self._extract_manifest(self.manifest_str, sql) or self.manifest_str
+        )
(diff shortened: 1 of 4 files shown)


goldmedal (approved):
Thanks @grieve54706 LGTM

PR #1024, opened 2025-01-03
feat(ibis): Added BE Support for MySQL SSL Connection

ongdisheng:
### Description
This PR adds SSL configuration support for MySQL connections where two new fields, `ssl_mode` and `ssl_ca`, have been introduced to enable passing SSL-related parameters to PyMySQL from ibis under the hood.

### Related Issue
https://github.com/Canner/WrenAI/issues/886

## Summary by CodeRabbit

- **New Features**
	- Added optional SSL configuration options for MySQL connections.
	- Introduced SSL mode settings (Disabled, Enabled, Verify CA).
	- Enhanced connection security with flexible SSL context management.
	- New testing capabilities for various SSL configurations in MySQL connections.
- **Bug Fixes**
	- Improved error handling for invalid SSL modes during MySQL connection attempts.

### DIFF
--- a/ibis-server/app/model/data_source.py
+++ b/ibis-server/app/model/data_source.py
@@ -29,14 +29,9 @@ from app.model import (
     QuerySnowflakeDTO,
     QueryTrinoDTO,
     SnowflakeConnectionInfo,
+    SSLMode,
     TrinoConnectionInfo,
 )
-
-
-class SSLMode(str, Enum):
-    DISABLE = "Disable"
-    REQUIRE = "Require"
-    VERIFY_CA = "Verify CA"
 
 
 class DataSource(StrEnum):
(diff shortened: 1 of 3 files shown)


ibis-server/app/model/data_source.py
```diff
@@ -30,6 +32,12 @@
 )
 
 
+class SSLMode(str, Enum):
+    DISABLE = "Disable"
+    REQUIRE = "Require"
+    VERIFY_CA = "Verify CA"
```
goldmedal, line 38:
```suggestion
    DISABLED = "disabled"
    ENABLED = "enabled"
    VERIFY_CA = "verify_ca"
```
I prefer to rename `require` -> `enabled`.
We can use the snake case for the enum value. The request body could be
```
{
      "sslMode": "verify_ca",
      "sslCA": "xxx",
}
```

goldmedal (commented):
Thanks for working it, @ongdisheng. Overall looks good.
I left some comments for the enum value. I wonder if we can add some tests to cover the connection information in `test_mysql`?
Maybe we can assert the error message when accepting `enable` or `verify_ca`. (the test container doesn't enable SSL by default.)

ibis-server/app/model/data_source.py
ongdisheng: Hi @goldmedal, I have tried adding a test for `ssl_mode=enabled` as shown below:
```python
async def test_connection_ssl_enabled(client, mysql: MySqlContainer):
    connection_info = _to_connection_info(mysql)
    connection_info["ssl_mode"] = "enabled"
    response = await client.post(
        url=f"{base_url}/metadata/version",
        json={"connectionInfo": connection_info},
    )
    assert response.status_code == 200
    assert response.text == '"8.0.40"'
```

However, the test passes, and I'm unsure if this is expected behavior due to MySQL server's default SSL configuration [ref: [link](https://dzone.com/articles/configuring-and-managing-ssl-on-your-mysql-server)].

> By default, MySQL server always installs and enables SSL configuration. However, it is not enforced that clients connect using SSL. Clients can choose to connect with or without SSL as the server allows both types of connections.

Your help is very much appreciated!

goldmedal (approved):
Thanks @ongdisheng. Overall looks good to me. 👍 
Sorry for being so late.

PR #1067, opened 2025-02-21
feat(ibis): Add Oracle connector

douenergy:
# Description
- Introduce Oracle datasource connector.
- The usage is the same as other datasource connector.

## Summary by CodeRabbit

- **New Features**
  - Introduced Oracle database integration to enable connection configuration, query execution, and metadata retrieval for Oracle systems.
- **Tests**
  - Added comprehensive tests to validate Oracle connectivity, query handling, and error scenarios.
- **Documentation**
  - Updated guides with clear instructions and examples to extend connection capabilities.
- **Chores**
  - Refined dependency management and tooling enhancements to support Oracle functionalities.


goldmedal:
@douenergy  You need to add `OracleConnectionInfo` to 
https://github.com/Canner/wren-engine/blob/90cd416275848b238c92003115ee6388de9a10e5/ibis-server/app/model/__init__.py#L202

We will verify the DTO when got a request
https://github.com/Canner/wren-engine/blob/90cd416275848b238c92003115ee6388de9a10e5/ibis-server/app/routers/v2/connector.py#L30

That's why the connection info way can't work.

ibis-server/tests/routers/v2/connector/test_oracle.py
```diff
@@ -0,0 +1,134 @@
...
+    assert result["dtypes"] is not None
+
+
+def _to_connection_info(oracle: OracleDbContainer):
+    return {
+        "host": oracle.get_container_host_ip(),
+        "port": oracle.get_exposed_port(oracle.port),
+        "user": "SYSTEM",
+        "password": "Oracle123",
+        "service": "FREEPDB1",
```
goldmedal, line 128:
```suggestion
        "database": "FREEPDB1",
```
This key name should match the definition in `OracleConnectionInfo`.

ibis-server/tests/routers/v2/connector/test_oracle.py
```diff
@@ -0,0 +1,134 @@
...
+@pytest.fixture(scope="module")
+def oracle(request) -> OracleDbContainer:
+    oracle = OracleDbContainer(
+        "gvenzl/oracle-free:23.6-slim-faststart", oracle_password="Oracle123"
+    )
+
+    oracle.start()
+
+    host = oracle.get_container_host_ip()
+    port = oracle.get_exposed_port(1521)
```
goldmedal, line 81:
```suggestion
    port = oracle.get_exposed_port(oracle.port)
```
I prefer to use the original port to avoid the potential port conflict.

ibis-server/tests/routers/v2/connector/test_oracle.py
```diff
@@ -0,0 +1,407 @@
...
+        url=f"{base_url}/metadata/tables",
+        json={"connectionInfo": connection_info},
+    )
+    assert response.status_code == 200
+
+    result = response.json()
+    # Use case-insensitive filtering since Oracle returns "SYSTEM.ORDERS"
+    result = next(filter(lambda x: x["name"].lower() == "system.orders", result))
+
+    assert result["name"].lower() == "system.orders"
```
goldmedal, line 354:
```suggestion
    result = next(filter(lambda x: x["name"] == "SYSTEM.ORDERS", result))
    assert result is not None
```
We can just match the upper case name and assert if it's None.

ibis-server/tests/routers/v2/connector/test_oracle.py
```diff
@@ -0,0 +1,407 @@
...
+
+async def test_metadata_db_version(client, oracle: OracleDbContainer):
+    connection_info = _to_connection_info(oracle)
+    response = await client.post(
+        url=f"{base_url}/metadata/version",
+        json={"connectionInfo": connection_info},
+    )
+    assert response.status_code == 200
+    assert "23.0" in response.text
+
```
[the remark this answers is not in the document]
goldmedal: It's interesting. @douenergy, should we care about the LOB type? maybe add a test for it?

ibis-server/app/model/metadata/oracle.py
```diff
@@ -0,0 +1,223 @@
...
+                    name=row["COLUMN_NAME"],
+                    type=self._transform_column_type(row["DATA_TYPE"]),
+                    notNull=row["IS_NULLABLE"] == "N",
+                    description=row["COLUMN_COMMENT"],
+                    properties=None,
+                )
+            )
+            # If the row indicates a primary key, set it.
+            if row.get("IS_PK", "NO") == "YES":
+                unique_tables[schema_table].primaryKey = row["COLUMN_NAME"]
```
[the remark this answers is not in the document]
goldmedal: @douenergy It makes sense to me. Could you check it? It's better to have a test to cover this case.

douenergy:
Thanks @goldmedal Jax for reviewing! I addressed most of the comments. 

But I don't think we need to explicitly add a LOB test, since we already define blob_column in the MDL, which is already a LOB column.

LOB is actually a logical type in Oracle, and BLOB is the physical type. You can see this in the Oracle documentation:
https://docs.oracle.com/en/database/oracle/oracle-database/23/adlob/LOB-classifications.html

goldmedal (approved):
Thanks @douenergy. look good 👍

PR #1068, opened 2025-02-24
fix(bigquery): correctly format the scientific notation decimal

douenergy:
Fixes formatting issues in BigQuery when handling scientific notation decimals, ensuring accurate representation


douenergy:
@goldmedal  We still have BigQuery secret issue, but all test include bigquery pass on my local. So i think it is ready for reveiw.

goldmedal (approved):
Thanks @douenergy, it looks good to me.

PR #1083, opened 2025-03-07
fix(engine): exclude the CTE table when source node checking

goldmedal:
# Description
Fix the following SQL syntax
```sql
select count(*) from (with orders_custkey as (select custkey from Orders) select * from orders_custkey) 
```
We should exclude the CTE table when we check if a table is a source table.

## Summary by CodeRabbit

- **New Features**
  - Improved SQL query processing by enhancing the resolution of table names for more accurate rewriting.
  
- **Tests**
  - Added a new test to validate SQL generation for counting queries involving nested common table expressions.


douenergy (approved):
Thanks @goldmedal 
LGTM 👍

PR #1088, opened 2025-03-12
feat(ibis): persistent query cache 

douenergy:
* Implemented a persistent query cache system that saves results to disk as Parquet files (`/tmp/wren-engine/`)

* Added an enable_cache parameter to the query endpoint to allow clients to opt-in to cache functionality.

* Integrated cache hit/miss tracking via an `X-Cache-Hit` response header for monitoring and debugging.
   I prefer header instead of a json body field. Because clean separation of metadata from data, follows HTTP convention for metadata, doesn't modify the existing response structure.

## Summary by CodeRabbit

- **New Features**
  - Introduced query caching to improve performance by reducing redundant executions.
  - Endpoints now support an optional caching parameter and include response headers that indicate cache usage.

- **Tests**
  - Added asynchronous tests to verify the caching behavior by comparing results on initial and subsequent requests for both v2 and v3 endpoints.


ibis-server/app/routers/v2/connector.py
```diff
@@ -44,16 +50,47 @@ async def query(
...
+        if enable_cache:
+            cached_result = query_cache_manager.get(
+                str(data_source), dto.sql, dto.connection_info
+            )
+            cache_hit = cached_result is not None
+
+        # Cache Hit !
+        if cached_result is not None:
+            response = ORJSONResponse(to_json(cached_result))
+            response.headers["X-Cache-Hit"] = str(cache_hit).lower()
```
goldmedal, line 66: Instead of a header field, how about using the query parameter? Just like dry-run, the client can invoke this API like
```
post .../v2/bigquery/query?cahceHit=true
```

goldmedal (commented):
Thanks @douenergy. Overall, it looks great, but something is missing. 
- This PR only implements the cache for the v2 API. However, the v3 API also requires it.
- As per our offline discussion yesterday, we want to make the cache storing be async. However, I think we don't need to do that in this PR. We can have another follow-up PR for the async thing. If you want to do that here, it's also ok for me.

douenergy:
@goldmedal  I have addressed all the comments including

1. Modified the tracking span name (appended cache_enable)
2. Added tracing for cache hit
3. Added v3 cache
4. Used OpenDAL to unify the FS interface
5. Moved test cases from separate files to Postgres test
6. Used HTTP query params instead of body params
7. For dry run queries, we don't care about the cache now - we just do what dry runs do because dry runs are just rewrites compared to execution, which is cheaper (I think this is a good idea)

ibis-server/app/routers/v2/connector.py
```diff
@@ -59,10 +69,43 @@ async def query(
...
-        return ORJSONResponse(to_json(connector.query(rewritten_sql, limit=limit)))
+            dry_response = Response(status_code=204)
+            dry_response.headers["X-Cache-Hit"] = "true"
+            return dry_response
+
+        # Not a dry run
+        # Check if the query is cached
+        cached_result = None
+        cache_hit = False
+        enable_cache = cache_enable
```
goldmedal, line 86:
```suggestion
```
I believe it's a redundant code.

ibis-server/tests/routers/v3/connector/postgres/test_query.py
```diff
@@ -136,6 +136,40 @@ async def test_query(client, manifest_str, connection_info):
     }
 
 
+async def test_query_with_cache(client, manifest_str, connection_info):
```
goldmedal, line 139: I think it's good to add the test in `tests/routers/v3/connector/postgres/test_fallback_v2.py`. To cover the behavior for fallback.

ibis-server/app/routers/v3/connector.py
```diff
@@ -54,15 +60,50 @@ async def query(
...
-            return ORJSONResponse(to_json(connector.query(rewritten_sql, limit=limit)))
+                dry_response = Response(status_code=204)
+                dry_response.headers["X-Cache-Hit"] = "true"
+                return dry_response
+
+            # Not a dry run
+            # Check if the query is cached
+            cached_result = None
+            cache_hit = False
+            enable_cache = cache_enable
```
goldmedal, line 71: It's also redundant.

ibis-server/app/routers/v2/connector.py
```diff
@@ -59,10 +69,43 @@ async def query(
...
+
+        # First check if the query is a dry run
+        # If it is dry run.
+        # We don't need to check query cache
         if dry_run:
             connector.dry_run(rewritten_sql)
-            return Response(status_code=204)
-        return ORJSONResponse(to_json(connector.query(rewritten_sql, limit=limit)))
+            dry_response = Response(status_code=204)
+            dry_response.headers["X-Cache-Hit"] = "true"
```
goldmedal, line 79: It's not right. Dry run success doesn't mean the cache hit. It only validates the SQL; we shouldn't return the cache hit header.

goldmedal (approved):
Thanks @douenergy. 👍

PR #1091, opened 2025-03-14
fix(ibis): Enforcing MySQL utf8mb4 charset in connections

douenergy:
Fix https://github.com/Canner/WrenAI/issues/1374

## Summary by CodeRabbit

- **New Features**
  - Improved MySQL connection handling now defaults to a robust UTF-8 variant, ensuring more reliable support for extended characters.
  
- **Tests**
  - Introduced a new test validating the handling of Chinese column names and data, reinforcing data integrity and seamless operation with diverse character sets.


ibis-server/tests/routers/v2/connector/test_mysql.py
```diff
@@ -475,6 +475,80 @@ async def test_connection_valid_ssl_mode(client, mysql_ssl_off: MySqlContainer):
     assert response.text == '"8.0.40"'
 
 
+# MySQL specific chinese tests
+# So we dont mess up with the other tests's connection
```
goldmedal, line 479:
```suggestion
```
It's a PR comment, not a codebase comment. I prefer to remove it.

ibis-server/tests/routers/v2/connector/test_mysql.py
```diff
@@ -475,6 +475,80 @@ async def test_connection_valid_ssl_mode(client, mysql_ssl_off: MySqlContainer):
...
+async def test_chinese_column_name_and_data(client):
```
goldmedal, line 480: I found that this case doesn't cover the change
```
        kwargs.setdefault("charset", "utf8mb4")
```
I tried to remove the charset setting. The test is still passed.
Could you double-check it?

douenergy:
The root issue stems from the Ibis server image utilizing the libmysqlclient-dev MySQL driver, which differs  from the driver implementation on macOS. I've successfully built and deployed the Ibis server image locally and conducted testing with Wren AI. The results confirm full UTF-8 compatibility.

goldmedal (approved):
Thanks @douenergy LGTM

PR #1100, opened 2025-03-19
chore(ibis): add the fallback mechanism for v3 API

goldmedal:
# Description
- Wren AI is migrating to v3 API. However, the engine side keeps a fallback mechanism for backward supporting the legacy SQL syntax. 
- We also log the event to the stdout and encourage the user to report the error.


douenergy (approved):
LGTM ! Thanks @goldmedal

goldmedal:
Thanks @douenergy 👍

PR #1104, opened 2025-03-25
fix(core): handle the unnest table factor for the specific dialect

goldmedal:
# Description
`unnest_as_table_factor` is required for the specific dialect only (e.g. BigQuery). We should enable it for BigQuery only.

### DIFF
--- a/wren-core/core/src/mdl/mod.rs
+++ b/wren-core/core/src/mdl/mod.rs
@@ -449,6 +449,7 @@ mod test {
     use datafusion::config::ConfigOptions;
     use datafusion::prelude::{SessionConfig, SessionContext};
     use datafusion::sql::unparser::plan_to_sql;
+    use wren_core_base::mdl::DataSource;
 
     #[test]
     fn test_sync_transform() -> Result<()> {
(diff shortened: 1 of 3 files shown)


wren-core/core/src/mdl/mod.rs
```diff
@@ -976,6 +977,25 @@ mod test {
...
+        let actual = transform_sql_with_ctx(&ctx, analyzed_mdl, &[], sql).await?;
+        assert_eq!(actual, "SELECT \"UNNEST(make_array(Int64(1),Int64(2),Int64(3)))\" FROM (SELECT UNNEST([1, 2, 3]) AS \"UNNEST(make_array(Int64(1),Int64(2),Int64(3)))\")");
+
+        let manifest = ManifestBuilder::new()
+            .data_source(DataSource::BigQuery)
+            .build();
+        let analyzed_mdl = Arc::new(AnalyzedWrenMDL::analyze(manifest)?);
+        let sql = "select * from unnest([1, 2, 3])";
+        let actual = transform_sql_with_ctx(&ctx, analyzed_mdl, &[], sql).await?;
+        assert_eq!(actual, "SELECT \"UNNEST(make_array(Int64(1),Int64(2),Int64(3)))\" FROM UNNEST([1, 2, 3])");
```
goldmedal, line 995: It's not a valid SQL for BigQuery actually. See https://github.com/apache/datafusion/issues/15233

douenergy (approved):
LTGM , Thanks @goldmedal

Issue #1105, opened 2025-03-25
SQL primary key detenction and composite key management


yalexbadan:
**Bug description**
Trying the product using the compose defined in [docker-compose.yaml](https://github.com/Canner/WrenAI/blob/main/docker/docker-compose.yaml), I noticed that primary keys was not detected for PostgreSQL connections.
Investigating on [postgres.py](https://github.com/Canner/wren-engine/blob/main/ibis-server/app/model/metadata/postgres.py#L19) I noticed that primary keys are not manged (same for Oracle), while the implementations for MySql and MsSql consider only the last detected column in composite primary key:
```python
def get_table_list(self) -> list[Table]:
  ...      
  for row in response:
    ...   
    # if column is primary key
    if row["column_key"] == "PRI":
      unique_tables[schema_table].primaryKey = row["column_name"]
```
The if at the end of for loops takes only the last detected column.
I don't know Trino nor Snowflakes to determine if both implementations are affected from the same problem.

**How to Reproduce**
Run the ibis-server, and try to invoke the `/v2/connector/{dbms}/metadata/tables` endpoints for all the connectors.

**Expected behavior**
In my proposal, the `primaryKey` field of class `Table` ([dto.py](https://github.com/Canner/wren-engine/blob/main/ibis-server/app/model/metadata/dto.py#L74)) should be a list of strings, with all fields belonging to the key.

**Desktop (please complete the following information):**
Test were perfomed using only docker images

**Wren AI Information**
- Versions:
  - `WREN_PRODUCT_VERSION=0.17.0`
  - `WREN_ENGINE_VERSION=0.14.8`
  - `WREN_AI_SERVICE_VERSION=0.17.1`
  - `IBIS_SERVER_VERSION=0.14.8`
  - `WREN_UI_VERSION=0.22.1`
  - `WREN_BOOTSTRAP_VERSION=0.1.5`

PR #1106, opened 2025-03-25
fix(ibis): added primary key management on postgres and updated mssql and mysql to manage composite keys

yalexbadan:
In this PR I added development proposal related to [issue 1105](https://github.com/Canner/wren-engine/issues/1105) (only about PostgreSql and composite key management).

Further investigating on `wren-ui`, I fear that the same fix should be popagated to table ER diagram page, but I don't have enought competence in font-end technologies to determine which classes should be updated.
I suppose that at least in the [__types__.ts](https://github.com/Canner/WrenAI/blob/main/wren-ui/src/apollo/client/graphql/__types__.ts), `primaryKey?: InputMaybe<Scalars['String']>` fields should become `primaryKey?: InputMaybe<Array<Scalars['String']>>`, and in [ModelForm.tsx](https://github.com/Canner/WrenAI/blob/main/wren-ui/src/components/pages/modeling/form/ModelForm.tsx), the selection should accempt multiple fields, but I wasn't able to figure out other updates.

## Summary by CodeRabbit

- **New Features**
  - Improved detection and representation of primary keys in PostgreSQL table metadata, including support for composite primary keys.

- **Bug Fixes**
  - Corrected SQL join condition in constraint retrieval for PostgreSQL metadata.

- **Tests**
  - Updated test manifests to standardize the format of the primaryKey field.
  - Minor indentation corrections in Java test files for improved readability.

- **Chores**
  - Updated .gitignore to exclude development container and VS Code configuration files.


goldmedal:
Thanks for contributing, @yalexbadan. I trigger the CI flow. I'll take a look at it after all workflows have passed.

goldmedal:
For the PR title, you need to follow the format of [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary).

goldmedal:
@yalexbadan It seems to be a code style issue. You can install [just](https://github.com/casey/just), then run the following command in the `ibis-server` folder to fix it.
```
just fmt 
```

goldmedal:
It seems some CI is queued and will never be released 🤔 . I'll try to close it and reopen the PR.

goldmedal:
The testing for `trino` and `clikchouse` are impacted. I guess it's related to the change in `dto.py`. You can see the detail in the [CI log](https://github.com/Canner/wren-engine/actions/runs/14126898111/job/39577905124?pr=1106).
```
tests/routers/v2/connector/test_clickhouse.py:514
tests/routers/v2/connector/test_trino.py:362
```

goldmedal:
@yalexbadan I think you shouldn't chagne the primary key in MDL to a list. The rewriter isn't support a model with multiple primary key.

For the composite key case, the right way is creating a column to concat the key. Then, use the new column as the primiary key.
For example:
```
model: {
   name: "model_1",
   columns: [
      {
         "name": "col1",
        ...
      },
      {
           "name": "col2",
          ...
      },
      {
            "name": "col3",
             "exrpession": "concat(col1, col2)",
              ...
       }
   ],
    "primaryKey": "col3"
}
```

yalexbadan:
> @yalexbadan I think you shouldn't chagne the primary key in MDL to a list. The rewriter doesn't support a model with multiple primary key.
> 
> For the composite key case, the right way is creating a column to concat the key. Then, use the new column as the primiary key. For example:
> 
> ```
> model: {
>    name: "model_1",
>    columns: [
>       {
>          "name": "col1",
>         ...
>       },
>       {
>            "name": "col2",
>           ...
>       },
>       {
>             "name": "col3",
>              "exrpession": "concat(col1, col2)",
>               ...
>        }
>    ],
>     "primaryKey": "col3"
> }
> ```

I'm trying to figure out from the code what I should change, but I'm not sure what you're referring to.
Are you referring to `mdl.schema.json` or `model.metadata.*`?

goldmedal:
The code of the java engine and rust engine shouldn't be changed:
`wren-base`, `wren-core`, `wren-core-base`, `wren-core-py`, `wren-tests`.

This change is only related to `test_metadata_*` in `tests/routers/v2/connector`

yalexbadan:
I re-read the code and re-analysed the solution I am trying to develop, but I am still not clear about your proposal.
Basically, it would be to remove what I have done completely, and simply create a "custom" field at `metadata` level, and then add that custom field in place of the `primaryKey`. It's not clear to me why intervene so massively on all components at this point?

goldmedal:
>  It's not clear to me why intervene so massively on all components at this point?

Because the wren-engine supports only one column for the primary key currently, it's related to the SQL rewriting of another feature called `calculated field` and `relationship`. It's not so simple.
Instead, we can support getting the primary key columns through `metadata` API (what you did in e159f5ab457aa8e8045ff46de9e30dd8dacd10cb) and build the composite key as I mentioned above:
```
model: {
   name: "model_1",
   columns: [
      {
         "name": "col1",
        ...
      },
      {
           "name": "col2",
          ...
      },
      {
            "name": "col3",
             "exrpession": "concat(col1, col2)",
              ...
       }
   ],
    "primaryKey": "col3"
}
```
We don't need to change the structure of the `model` in java or rust engine.

yalexbadan:
Testing the application using the ui, I got this error:
```
{"errors":[{"locations":[{"line":2,"column":3}],"path":["saveTables"],"message":"Cannot read properties of undefined (reading 'replace')","extensions":{"code":"INTERNAL_SERVER_ERROR","message":"Cannot read properties of undefined (reading 'replace')","shortMessage":"Internal server error"}}],"data":null}
```
I'm trying to figure out possibile regressions, but I'm not expert in front-end tecnologies, so I'm not sure if I will be able to identify where the error is thrown

ibis-server/app/model/metadata/postgres.py
```diff
@@ -71,15 +81,38 @@ def get_table_list(self) -> list[Table]:
...
+                unique_tables[schema_table].primaryKey = columns[0].name
+            else:
+                composite_primary_key_column = Column(
+                    name="composed_primary_key",
+                    type="json",
+                    notNull=True,
+                    description=f"Composed primary key based on fields: {[col.name for col in columns]}",
                     properties=None,
+                    nestedColumns=columns,
                 )
```
goldmedal, line 112: Is it possible to have multiple primary keys in the same table? IMO, the postgres only allows the table has only one primary key. Given a table `customer`:
```
test=# alter table customer add constraint cky primary key (c_custkey);
ALTER TABLE
test=# alter table customer add constraint cky primary key (c_name);
ERROR:  multiple primary keys for table "customer" are not allowed
```

ibis-server/app/model/metadata/oracle.py
```diff
@@ -101,6 +101,8 @@ def get_table_list(self) -> list[Table]:
                 )
             )
 
+            # TODO: manage primary key
+
```
goldmedal, line 105: What's the comment propose? Could you explain more?

goldmedal (commented):
Thanks @yalexbadan 
The primary key part is pretty close, but I have some concerns about the composite key. IMO, the postgres table is only allowed to have one primary key. Do you have any real case for the composite key?
By the way, I would like to add some tests for this PR if you don't mind me push some commits to your branch.

PR #1131, opened 2025-04-02
fix(core): simplify the available function lists

goldmedal:
# Description
To avoid making the list too large for the LLM promotion, we only return the function name, description, and the function type. 

## Summary by CodeRabbit

- **Refactor**
  - Streamlined the function registration and metadata retrieval process to return only essential details (name, type, and an optional description), omitting previous parameter and return type specifics.
  - Adjusted the reported count of available functions for a clearer, more accurate overview.

- **Tests**
  - Updated test validations across multiple connectors to align with the simplified function metadata.

### DIFF
--- a/ibis-server/tests/conftest.py
+++ b/ibis-server/tests/conftest.py
@@ -11,7 +11,7 @@ def file_path(path: str) -> str:
     return os.path.join(os.path.dirname(__file__), path)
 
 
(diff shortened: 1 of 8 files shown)


douenergy (approved):
LGTM ! 
Thanks @goldmedal

PR #1138, opened 2025-04-08
feat(ibis): override query cache

douenergy:
- Add creation timestamp as part of the query cache file name
- When cache hit occurs, also return the cache creation timestamp
- Add an override parameter so that when cache is enabled, you can still force a query to the data source and then cache the result
- By default, the override parameter is set to false

## Summary by CodeRabbit

- **New Features**
  - Introduced an option to bypass cached query results using an override parameter, providing greater control over data freshness.
  - Enhanced responses with additional headers that indicate cache creation timestamps and when an override is applied.
  - Improved overall cache management to ensure more consistent retrieval of up-to-date query results.

- **Tests**
  - Added validations to ensure the new cache override functionality and response metadata work as expected.
  - Expanded testing coverage for cache behavior, particularly focusing on scenarios involving cache overrides.


goldmedal (approved):
Thanks @douenergy. It looks great 👍

PR #1145, opened 2025-04-10
fix(core): expand the wildcard before Wren rewrite rules

goldmedal:
The result of the following SQL is wrong:
```
select * from "Order_items" where "Order_id" in ('03c83b31dbc387f83f1b5579b53182fb', '08cbb1d4cd574b126569b208fd4b26ea')
```
The generated SQL will miss the columns because Wren Core only collects the column in the filter.
Since the `ExpandWildcardRule` will be removed from the latest DataFusion branch. To align this change, I apply the rule before all rules. This behavior is similar to expanding the wildcard in the logical plan builder.

## Summary by CodeRabbit

- **Refactor**
  - Updated internal processing to prioritize key transformation steps, ensuring consistent behavior.

- **Tests**
  - Enhanced SQL query validations with additional nested query checks.
  - Introduced a new test case to validate wildcard filtering.
  - Added new queries in the test suite to verify accurate data retrieval for order items.
  - Modified expected outputs in existing test cases to reflect new SQL structures and data types.

### DIFF
--- a/ibis-server/tests/routers/v3/connector/postgres/test_query.py
+++ b/ibis-server/tests/routers/v3/connector/postgres/test_query.py
@@ -111,16 +111,16 @@ async def test_query(client, manifest_str, connection_info):
     assert len(result["columns"]) == 10
     assert len(result["data"]) == 1
     assert result["data"][0] == [
+        1,
+        370,
+        "O",
+        "172799.49",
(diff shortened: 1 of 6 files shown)


douenergy (approved):
Thanks @goldmedal

PR #1152, opened 2025-04-16
feat(substitute): add default header to model substitute

douenergy:
This PR adds support for two optional HTTP header `x-user-catalog` and `x-user-schema` in the following endpoints:

POST `/v2/connector/{data_source}/model-substitute`

POST `/v3/connector/{data_source}/model-substitute`

These headers allow clients to explicitly specify the catalog and schema context when performing model substitution. 

## Summary by CodeRabbit

- **New Features**
  - Added support for specifying catalog and schema via HTTP headers across multiple endpoints, including model substitution.

- **Bug Fixes**
  - Ensured consistent model substitution behavior regardless of whether catalog and schema are included in SQL queries or provided via headers.
  - Improved error handling when only one of the required headers (`x-user-catalog` or `x-user-schema`) is provided.

- **Tests**
  - Expanded test coverage to validate model substitution with catalog and schema in both SQL and headers, covering success and error cases.

### DIFF
--- a/ibis-server/app/mdl/substitute.py
+++ b/ibis-server/app/mdl/substitute.py
@@ -1,4 +1,3 @@
-from loguru import logger
 from opentelemetry import trace
 from sqlglot import exp, parse_one
 from sqlglot.optimizer.scope import build_scope
@@ -42,25 +41,31 @@ class ModelSubstitute:
         def key(model):
             table_ref = model["tableReference"]
 
-            if not table_ref.get("catalog") and not table_ref.get("schema"):
-                logger.debug(
-                    "Try to substitute a tableReference has empty catalog and empty schema"
-                )
-
-            return f"{table_ref.get('catalog', '')}.{table_ref.get('schema', '')}.{table_ref.get('table', '')}"
+            # fully qualified  catalog.schema.table
+            if table_ref.get("catalog") and table_ref.get("schema"):
+                return f"{table_ref.get('catalog', '')}.{table_ref.get('schema', '')}.{table_ref.get('table', '')}"
+            # schema.table
+            elif table_ref.get("schema"):
+                return f"{table_ref.get('schema', '')}.{table_ref.get('table', '')}"
+            # table
+            else:
+                return table_ref.get("table", "")
 
         return {key(model): model for model in models if "tableReference" in model}
 
     def _find_model(self, source: exp.Table) -> dict | None:
-        if source.catalog and source.db:
+        # Determine catalog
+        if source.catalog:
             catalog = source.catalog
-            schema = source.db
-        elif self.headers is not None:
-            catalog = self.headers.get("x-user-catalog")
-            schema = self.headers.get("x-user-schema")
         else:
-            catalog = ""
-            schema = ""
+            catalog = self.headers.get("x-user-catalog", "") if self.headers else ""
+
+        # Determine schema
+        if source.db:
+            schema = source.db
+        else:
+            schema = self.headers.get("x-user-schema", "") if self.headers else ""
+
         table = source.name
         return self.model_dict.get(f"{catalog}.{schema}.{table}", None)
 
--- a/ibis-server/app/mdl/substitute.py
+++ b/ibis-server/app/mdl/substitute.py
@@ -10,10 +10,11 @@ tracer = trace.get_tracer(__name__)
 
 
 class ModelSubstitute:
-    def __init__(self, data_source: DataSource, manifest_str: str):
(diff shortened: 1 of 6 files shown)


ibis-server/app/routers/v2/connector.py
```diff
@@ -271,16 +271,17 @@ async def dry_plan_for_data_source(
 async def model_substitute(
     data_source: DataSource,
     dto: TranspileDTO,
+    request: Request,
```
goldmedal, line 274: I think `headers` has been injected at L276. Why do we need to inject `request` again?

ibis-server/tests/routers/v2/connector/test_postgres.py
```diff
@@ -710,12 +711,29 @@ async def test_dry_plan(client, manifest_str):
...
+            "sql": 'SELECT * FROM "test"."public"."orders"',
+        },
+    )
+    assert response.status_code == 200
+    assert (
+        response.text
+        == '"SELECT * FROM \\"my_catalog\\".\\"my_schema\\".\\"Orders\\" AS \\"orders\\""'
+    )
+
+    # Test without catalog and schema in SQL but in headers(x-user-xxx)
```
goldmedal, line 729: Could you add a test for the case where the header only has `catalog` or only has `schema`?

goldmedal (commented):
Thanks @douenergy. It looks nice. However, I think we need to enhance the tests. I left some comments.
By the way, I found another potential issue at `_build_model_dict`.
This method is only used for the full qualified table reference. If a table reference isn't qualified, the hash key would be wrong.

ibis-server/tests/routers/v2/connector/test_postgres.py
```diff
@@ -724,19 +726,93 @@ async def test_model_substitute(client, manifest_str, postgres: PostgresContaine
...
+            "sql": 'SELECT * FROM "orders"',
+        },
+    )
+    assert response.status_code == 200
+    assert (
+        response.text
+        == '"SELECT * FROM \\"my_catalog\\".\\"my_schema\\".\\"Orders\\" AS \\"orders\\""'
+    )
+
+    # Test only have x-user-catalog
```
douenergy, line 745: I added 3 test cases that only include partial headers in v2

ibis-server/tests/routers/v3/connector/postgres/test_model_substitute.py
```diff
@@ -43,16 +45,102 @@ async def test_model_substitute(client, manifest_str, connection_info):
...
+            "sql": 'SELECT * FROM "orders"',
+        },
+    )
+    assert response.status_code == 200
+    assert (
+        response.text
+        == '"SELECT * FROM \\"my_catalog\\".\\"my_schema\\".\\"Orders\\" AS \\"orders\\""'
+    )
+
+    # Test only have x-user-catalog
```
douenergy, line 67: In v3

ibis-server/app/mdl/substitute.py
```diff
@@ -39,13 +41,26 @@ def substitute(self, sql: str, write: str | None = None) -> str:
     def _build_model_dict(models) -> dict:
         def key(model):
             table_ref = model["tableReference"]
+
+            if not table_ref.get("catalog") and not table_ref.get("schema"):
+                logger.debug(
+                    "Try to substitute a tableReference has empty catalog and empty schema"
+                )
```
goldmedal, line 48: I think we don't need this log. 🤔 
The problem is that if the table reference has only the table name, the return key is `..table`. However, it should be `table`. So I think we can change here:
If the table reference is fully qualified, the key is `catalog.schema.table`.
If the table reference is partially qualified, the key is `schema.table`.
If the table reference is unqualified, the key is `table`.

ibis-server/tests/routers/v2/connector/test_postgres.py
```diff
@@ -724,19 +726,93 @@ async def test_model_substitute(client, manifest_str, postgres: PostgresContaine
...
+    response = await client.post(
+        url=f"{base_url}/model-substitute",
+        headers={"x-user-catalog": "test"},
+        json={
+            "connectionInfo": connection_info,
+            "manifestStr": manifest_str,
+            "sql": 'SELECT * FROM "public"."orders"',
+        },
+    )
+    assert response.status_code == 422
```
goldmedal, line 767: Why does this case fail? If the default catalog is `test`, the table name will be `test.public.orders`. I think it can map to `"my_catalog"."my_schema"."Orders"`, right?

goldmedal (approved):
Look great!

PR #1157, opened 2025-04-21
feat(substitute): smart case model substitute 

douenergy:
In data sources like `Oracle` and `Snowflake`, schema and table names are case-insensitive. For example:

```sql 
SELECT * FROM SYSTEM.ORDERS
```

is equivalent to:

```sql
SELECT * FROM system.orders
```

However, in our current model definition, `TableReference` is case-sensitive by default.
This PR introduces logic to automatically match `TableReference` entries in a case-insensitive manner when resolving catalog and schema names, aligning with the behavior of these data sources.

## Summary by CodeRabbit

- **New Features**
  - Improved model substitution to support case-insensitive lookups, ensuring more robust handling of model names regardless of letter casing.

- **Tests**
  - Added tests to verify that model substitution works correctly with different casing in schema names, confirming case insensitivity in the substitution process.
  - Added validation to detect and reject ambiguous model names differing only by case, preventing substitution errors.

### DIFF
--- a/ibis-server/app/mdl/substitute.py
+++ b/ibis-server/app/mdl/substitute.py
@@ -39,7 +39,7 @@ class ModelSubstitute:
 
                 # if model name is ambiguous, raise an error
                 duplicate_keys = get_case_insensitive_duplicate_keys(self.model_dict)
-                if model is not None and key in duplicate_keys:
+                if model is not None and key.lower() in duplicate_keys:
                     raise SubstituteError(
                         f"Ambiguous model: found multiple matches for {source}"
                     )
@@ -125,9 +125,9 @@ def quote(s: str) -> str:
 def get_case_insensitive_duplicate_keys(d):
     key_map = defaultdict(list)
     for k in d:
-        key_map[k.lower()].append(k)
+        key_map[k.lower()].append(k.lower())
 
-    duplicates = [keys for keys in key_map.values() if len(keys) > 1]
+    duplicates = [key for keys in key_map.values() if len(keys) > 1 for key in keys]
     return duplicates
 
 
--- a/ibis-server/app/mdl/substitute.py
+++ b/ibis-server/app/mdl/substitute.py
@@ -30,7 +30,7 @@ class ModelSubstitute:
                     continue
 
                 model = self._find_model(source) or self._find_model(
-                    source, case_insensitive=False
+                    source, case_sensitive=False
                 )
                 if model is None:
                     raise SubstituteError(f"Model not found: {source}")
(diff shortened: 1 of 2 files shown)


ibis-server/app/mdl/substitute.py
```diff
@@ -53,7 +63,24 @@ def key(model):
...
+            if table_ref.get("catalog") and table_ref.get("schema"):
+                return f"{table_ref.get('catalog', '')}.{table_ref.get('schema', '')}.{table_ref.get('table', '')}".lower()
+            # schema.table
+            elif table_ref.get("schema"):
+                return f"{table_ref.get('schema', '')}.{table_ref.get('table', '')}".lower()
+            # table
+            else:
+                return table_ref.get("table", "").lower()
+
+        return {key(model): model for model in models if "tableReference" in model}
```
goldmedal, line 81: We can simplify here by extracting the key function like:
```python
    @staticmethod
    def _build_model_dict(models) -> dict:
        return {_build_key(model): model for model in models if "tableReference" in model}

    @staticmethod
    def _build_case_insensitive_model_dict(models) -> dict:
        return {_build_key(model).lower() : model for model in models if "tableReference" in model}

def _build_key(model):
    table_ref = model["tableReference"]

    # fully qualified  catalog.schema.table
    if table_ref.get("catalog") and table_ref.get("schema"):
        return f"{table_ref.get('catalog', '')}.{table_ref.get('schema', '')}.{table_ref.get('table', '')}"
    # schema.table
    elif table_ref.get("schema"):
        return f"{table_ref.get('schema', '')}.{table_ref.get('table', '')}"
    # table
    else:
        return table_ref.get("table", "")
```

ibis-server/app/mdl/substitute.py
```diff
@@ -68,14 +95,28 @@ def _find_model(self, source: exp.Table) -> dict | None:
...
-            return self.model_dict.get(f"{table}", None)
+            model_dict = self.model_dict_case_insensitive
+            # catalog and schema is not None and not empty string
+            if catalog and schema:
+                return model_dict.get(f"{catalog}.{schema}.{table}".lower(), None)
+            # catalog is not None and not empty string
+            elif schema:
+                return model_dict.get(f"{schema}.{table}".lower(), None)
+            else:
+                return model_dict.get(f"{table}".lower(), None)
```
goldmedal, line 119: We can simplify here like:
```python
        # catalog and schema is not None and not empty string
        if catalog and schema:  
            key = f"{catalog}.{schema}.{table}"
        # schema is not None and not empty string
        elif schema:
            key = f"{schema}.{table}"
        else:
            key = f"{table}"

        # Determine if case insensitive search is needed
        if case_sensitive:
            return self.model_dict.get(key, None)
        else:
            return self.model_dict_case_insensitive.get(key.lower(), None)
```

goldmedal (commented):
Thanks @douenergy. Some codebase can be simplified.
Then, I think we should add some test for the ambiguous model name case. For example, if we have two models, `Orders` and `orders`. Because the model name in Wren is case sensitive, if we input `ORDERS`, we may get two candidates. We should throw the ambiguous error for this case.

goldmedal (approved):
Thanks @douenergy. Look great 👍

PR #1162, opened 2025-04-25
feat(ibis): query cache with connection url  

douenergy:
We can support query cache from connection info using connection URL.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cache key generation to support both connection URLs and traditional connection info, ensuring correct caching behavior for different connection types.

- **Tests**
  - Added tests to verify query caching and cache override functionality when using connection URLs for PostgreSQL connections. These tests confirm correct cache hits, overrides, and consistent query results.

### DIFF
--- a/ibis-server/app/query_cache/__init__.py
+++ b/ibis-server/app/query_cache/__init__.py
@@ -7,6 +7,8 @@ import opendal
 from loguru import logger
 from opentelemetry import trace
 
+from app.model import ConnectionUrl
+
 tracer = trace.get_tracer(__name__)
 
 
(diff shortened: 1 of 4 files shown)


ibis-server/app/query_cache/__init__.py
```diff
@@ -70,13 +72,18 @@ def get_cache_file_timestamp(self, data_source: str, sql: str, info) -> int | No
...
+            key_parts = [data_source, sql, info.connection_url.get_secret_value()]
+        else:
+            key_parts = [
+                data_source,
+                sql,
+                info.host.get_secret_value(),
+                info.port.get_secret_value(),
+                info.user.get_secret_value(),
+            ]
+        logger.debug(f"Hash key components: {key_parts}")
```
goldmedal, line 86:
```suggestion
        logger.debug("Hash key components: {}", key_parts)
```
It's better to input `key_parts` as an argument.

goldmedal (commented):
Could you add the case for the fallback test? Others make sense to me.

goldmedal (approved):
Thanks @douenergy 👍

PR #1167, opened 2025-04-25
feat(ibis): query cache with bigquery info 

douenergy:
We can support query cache from connection info using bigquery connection info.

## Summary by CodeRabbit

- **New Features**
	- Improved query caching support for BigQuery connections, ensuring cache keys are generated using relevant connection details.
- **Tests**
	- Added tests to verify query caching behavior, including cache hits, cache overrides, and related response headers.


goldmedal (approved):
Thanks @douenergy  👍

PR #1180, opened 2025-05-05
fix(ibis): skip redundant operations when query cache is hit

douenergy:
- Skips establishing a new database connection if the result is already cached.

- Bypasses SQL rewriting logic when cached results are used.

Previously, even on a cache hit, the system still performed operations like opening a database connection and rewriting the SQL, which were redundant. This change improves efficiency and reduces latency during cache hits.

## Summary by CodeRabbit

- **Refactor**
  - Improved query caching for faster responses on cache hits with clearer cache-related headers.
  - Reduced redundant SQL rewriting and connector initialization by performing them only when needed.
  - Enhanced clarity and reliability in handling cache overrides, cache misses, and fallback disable scenarios.
  - Updated test behavior to ensure consistent successful responses when fallback is disabled, validating cached results remain accurate.

### DIFF
--- a/ibis-server/app/routers/v3/connector.py
+++ b/ibis-server/app/routers/v3/connector.py
@@ -69,12 +69,12 @@ async def query(
         name=span_name, kind=trace.SpanKind.SERVER, context=build_context(headers)
     ) as span:
         try:
-            sql = pushdown_limit(dto.sql, limit)
-            rewritten_sql = await Rewriter(
-                dto.manifest_str, data_source=data_source, experiment=True
-            ).rewrite(sql)
-            connector = Connector(data_source, dto.connection_info)
             if dry_run:
+                sql = pushdown_limit(dto.sql, limit)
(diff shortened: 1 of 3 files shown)


ibis-server/app/routers/v3/connector.py
```diff
@@ -88,61 +93,67 @@ async def query(
                     data_source, dto.sql, dto.connection_info
                 )
                 cache_hit = cached_result is not None
+            # case 1: cache hit read
+            # We also need to check `is_fallback_disable` because a previous request
+            # might have fallen back successfully  and created a cache.
+            # If the current request has fallback disabled, we should bypass the cache hit check.
```
goldmedal, line 99: I think it's wrong. If we always bypass the cache hit check, it means the cache never works for the request, which disables the fallback. IMO, it's fine to use the cache created by the previous fallback request.

goldmedal (approved):
Thanks @douenergy. LGTM 👍

PR #1188, opened 2025-05-06
feat(ibis): postgreSQL, PostGIS support for IBIS Server.

eren-ture:
### Inclusions

* Added an `ExtensionHandler` class to handle database extensions. Created the `postgis_handler` method to define columns as `geometry` or `geography`.

* Added new column types `GEOMETRY` and `GEOGRAPHY` to the `RustWrenEngineColumnType` enum to represent PostGIS-specific data types.

### Proof of Work
image
image

### Issue

When the frontend is asked to display the values, it obviously fails. I unfortunately don't know much about frontend development, ad cannot fix this issue.
image

## Summary by CodeRabbit

- **New Features**
  - Added support for recognizing and displaying PostGIS extension column types (GEOMETRY and GEOGRAPHY) in PostgreSQL metadata.
  - Metadata extraction now includes extension-specific columns, providing more accurate information for tables using supported PostgreSQL extensions.
  - Introduced spatial data handling and validation through PostGIS-enabled testing, ensuring accurate spatial query support.
  - Added new dependencies to support spatial data processing and integration.


chilijung:
This is very interesting use case! @wwwy3y3 @andreashimin take a closer look!

goldmedal:
Thanks @eren-ture for the contribution. It's awesome. 👍
However, it's better to add the corresponding test case for it.
You can add the test case in `ibis-server/tests/routers/v2/connector/test_postgres.py`
https://github.com/Canner/wren-engine/blob/c0f888d4e4b99003866ae212e7f5b9c9135992a0/ibis-server/tests/routers/v2/connector/test_postgres.py#L742

Maybe add a test case called `test_metadata_list_tables_with_postgis`.
I think you need to prepare the testing data by initializing the testing Postgres with PostGIS and try to list tables from it. 
We use `testcontainer` to create the testing PG when testing. You can refer to `postgres()` to see how we initialize a testing database. 
https://github.com/Canner/wren-engine/blob/c0f888d4e4b99003866ae212e7f5b9c9135992a0/ibis-server/tests/routers/v2/connector/test_postgres.py#L136

I expected you might implement another method called `postgres_gis()` for another instance.

ibis-server/app/model/metadata/postgres.py
```diff
@@ -11,6 +11,92 @@
...
+
+        # Update tables
+        for row in response:
+            table_name = f"{row['f_table_schema']}.{row['f_table_name']}"  # ? Might want to use a global `_format_postgres_compact_table_name` function.
+            table = tables[table_name]
+            for column in table.columns:
+                if column.name == row["column_name"]:
+                    column.type = self._transform_postgres_column_type(
+                        row["column_type"]
+                    )
```
goldmedal, line 82:
```suggestion
                    column.type = str(self._transform_postgres_column_type(
                        row["column_type"]
                    ))
```
I got the following message when testing:
```
  Expected `str` but got `RustWrenEngineColumnType` with value `<RustWrenEngineColumnType.GEOMETRY: 'GEOMETRY'>` - serialized value may not be as expected
```
I think it's better to cast to `str` explicitly to avoid the warning.

ibis-server/app/model/metadata/postgres.py
```diff
@@ -11,6 +11,92 @@
...
+                f_geography_column AS column_name,
+                'geography' AS column_type
+            FROM
+                {schema_name}.geography_columns;
+            """
+        response = self.connection.sql(sql).to_pandas().to_dict(orient="records")
+
+        # Update tables
+        for row in response:
+            table_name = f"{row['f_table_schema']}.{row['f_table_name']}"  # ? Might want to use a global `_format_postgres_compact_table_name` function.
```
goldmedal, line 76:
```suggestion
            # TODO: Might want to use a global `_format_postgres_compact_table_name` function.
            table_name = f"{row['f_table_schema']}.{row['f_table_name']}"
```
The comment is like a to-do enhancement. It's better to add the TODO  word for it.

goldmedal (commented):
Thanks @eren-ture, I think this PR is pretty close. I left some comments for it. As what I mentioned previously, it's good to add the test case for it. We can use [the Postgis image](https://hub.docker.com/r/postgis/postgis) to build the test cases. I think I can help you do this if you don't mind I push some commits to your branch.

eren-ture:
Hi @goldmedal,
I'm currently working on writing the tests. I'm more on the data side than software, but I'll probably get some commits out soon.

goldmedal:
> Hi @goldmedal, I'm currently working on writing the tests. I'm more on the data side than software, but I'll probably get some commits out soon.

Thanks. I'm looking forward to them. 👍

ibis-server/tests/routers/v2/connector/test_postgres.py
```diff
@@ -149,6 +150,19 @@ def postgres(request) -> PostgresContainer:
     return pg
 
 
+@pytest.fixture(scope="module")
+def postgis(request) -> PostgresContainer:
+    pg = PostgresContainer("postgis/postgis:16-3.5-alpine").start()
```
goldmedal, line 155: Because `postgis` only provides the amd64 images, I faced an error about the image not found on my M4 Mac. We can solve it by downloading the image manually:
```
docker pull --platform linux/amd64 postgis/postgis:16-3.5-alpine
```
I think it's better to add a comment to mention it.

goldmedal (commented):
Thanks @eren-ture. The CI failure is caused by the failure to install the required dependencies. I think we need to add the following dependency in `ibis-server/pyproject.toml` and update the lock file.
```
geopandas = "^1.0.1"
geoalchemy2 = "^0.17.1"
```

ibis-server/tests/routers/v2/connector/test_postgres.py
eren-ture: > Thanks @eren-ture. The CI failure is caused by the failure to install the required dependencies. I think we need to add the following dependency in `ibis-server/pyproject.toml` and update the lock file.
> 
> ```
> geopandas = "^1.0.1"
> geoalchemy2 = "^0.17.1"
> ```

Ohhh, yes. Completely slipped my mind.

> Because `postgis` only provides the amd64 images, I faced an error about the image not found on my M4 Mac. We can solve it by downloading the image manually:
> 
> ```
> docker pull --platform linux/amd64 postgis/postgis:16-3.5-alpine
> ```
> 
> I think it's better to add a comment to mention it.

👍

douenergy:
@eren-ture Thanks for all the hard work! Could you update the Poetry lock file? It looks like the CI failed because it needs to be updated. You can just run poetry lock.

eren-ture:
I rebased with main and ran poetry lock. Hopefully this should merge.

goldmedal (approved):
Thanks @eren-ture for this work 👍 It's awesome.

PR #1211, opened 2025-06-03
feat(core): implement column-level access control for the model

goldmedal:
# Description
This PR introduces the column-level access control(CLAC) for the model in the Rust implementation. CLAC is used to apply a rule to control whether a column is accessible according to the session properties.

# Spec
```json
  {
      "name": "c_name",
      "type": "varchar",
      "columnLevelAccessControl": {
          "name": "c_name_access",
          "requiredProperties": [
              {
                  "name": "session_level",
                  "required": False,
              }
          ],
          "operator": "EQUALS",
          "threshold": "1",
      },
  },
```
`columnLevelAccessControl` is an attribute of a column.
- `name`: The display name of the rule.
- `requiredProperties`: The required properties for this rule. It's the same with #1161.
    - It currently allows only one property for the CLAC.
- `operator`: The operator is used for the rule. The supported operators:
  - Equals
  - NotEquals
  - GreaterThan
  - LessThan
  - GreaterThanOrEquals
  - LessThanOrEquals
- `threshold`: The threshold of this rule.

# Example
The above case shows how to define a CLAC for the column `c_name`. If the session properties is
```
X-Wren-Variable-SESION_LEVEL: 2
```
Wren core will check the following formula:
```
X-Wren-Variable-SESION_LEVEL = 1
```
If it's passed, the column is accessible. Otherwise, the column won't be found.

## Summary by CodeRabbit

- **New Features**
  - Introduced column-level access control (CLAC), allowing columns to be conditionally included or excluded from query results based on session properties.
  - Session properties can now be passed and used to enforce access control at the column level.
- **Bug Fixes**
  - Improved error messaging for missing required session properties.
- **Tests**
  - Added tests to verify column-level access control behavior and session property handling.
- **Chores**
  - Updated dependencies for improved compatibility.
  - Deprecated and removed unused or legacy internal modules.


douenergy (approved):

douenergy:
Thanks @goldmedal

PR #1226, opened 2025-06-18
fix(core): disable EliminateLimit rule

goldmedal:
The EliminateLimit optimization rule will skip the plan if the plan is limited to 0. The unparser can't handle the generated plans. For example:
```
SELECT * FROM orders LIMIT 0
```
It will be unparsed to 
```
SELECT *
```

We should disable the rule to get the correct result.


douenergy:
Thanks @goldmedal

PR #1232, opened 2025-06-23
fix(ibis): Correctly filter Oracle tables by connected user's schema

### LINKED ISSUE #1175: Oracle Connector oracle.py t.owner="SYSTEM" user SYSTEM hardcoded, it can't bring other schema tables
**Describe the bug**
@douenergy it doesn't work, it always brings System. tables, it doesn't bring other tables like (HR schema ) reason : https://github.com/Canner/wren-engine/blob/main/ibis-server/app/model/metadata/oracle.py "SYSTEM" user hardcoded in this file: WHERE
t.owner = 'SYSTEM'
ORDER BY
t.table_name, c.column_id;. so if I have HR db, user will be HR, but above query doesn't bring tables because, t.owner="SYSTEM"

**To Reproduce**
Create as schema / user example HR (don't use System schema) and add one table.

**Expected behavior**
Metadata discovery should display this table, if you login with user HR

**Screenshots**

**Desktop (please complete the following information):**
- OS: [e.g. Windows, Linux, Mac, Mac Arm ...]

**Wren AI Information**
- Version: [e.g, 0.1.0]

**Additional context**
@douenergy it doesn't work, it always brings System. tables, it doesn't bring other tables like (HR schema ) reason : https://github.com/Canner/wren-engine/blob/main/ibis-server/app/model/metadata/oracle.py "SYSTEM" user hardcoded in this file: WHERE
t.owner = 'SYSTEM'
ORDER BY
t.table_name, c.column_id;. so if I have HR db, user will be HR, but above query doesn't bring tables because, t.owner="SYSTEM"

### LINKED ISSUE #1213: Oracle Data Source: Schema Selection Ignores Chosen Schema, Defaults to SYSTEM
**Describe the bug**
When using Oracle as a data source, regardless of which schema name is selected, the system always returns tables from the SYSTEM schema.

**To Reproduce**

Configure Wren AI engine to connect to an Oracle database.
Attempt to select a specific schema other than SYSTEM (e.g., SCOTT or a custom schema).
Observe the list of tables returned by the Wren AI engine.
Expected behavior
When a specific Oracle schema is inputed , the Wren AI engine should display only the tables from the inputed schema, not the SYSTEM schema.

OS:   Windows 10, Ubuntu 22.04 LTS
Wren Engine Version:  0.16.1
Oracle Database Version:  Oracle 12c

loaferxuan:
**Summary**
This PR addresses a bug in the `OracleMetadata.get_table_list` method where the table discovery was incorrectly limited to the SYSTEM schema. The fix ensures that the query correctly uses the connected user's schema to find tables.

**The Problem**
Currently, the SQL query within the `get_table_list` method contains a hardcoded WHERE clause: `t.owner = 'SYSTEM'`. This means that regardless of which user is configured in the connection, the system will only ever search for tables owned by SYSTEM. 

**Related Issues**
Closes #1213
Closes #1175
Closes Canner/WrenAI#1743

## Summary by CodeRabbit

- **Bug Fixes**
  - Table list now correctly displays tables from the current user's schema instead of always showing those from the 'SYSTEM' schema.

### DIFF
--- a/wren-core-legacy/pom.xml
+++ b/wren-core-legacy/pom.xml
@@ -10,7 +10,7 @@
 
     <groupId>io.wren</groupId>
     <artifactId>wren-root</artifactId>
-    <version>0.17.1</version>
+    <version>0.17.2-SNAPSHOT</version>
     <packaging>pom</packaging>
 
     <name>WrenEngine</name>
@@ -35,7 +35,7 @@
     </modules>
 
     <scm>
-        <tag>0.17.1</tag>
+        <tag>HEAD</tag>
     </scm>
 
     <distributionManagement>
--- a/wren-core-legacy/trino-parser/pom.xml
+++ b/wren-core-legacy/trino-parser/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>io.wren</groupId>
         <artifactId>wren-root</artifactId>
-        <version>0.17.1</version>
+        <version>0.17.2-SNAPSHOT</version>
         <relativePath>../pom.xml</relativePath>
     </parent>
 
--- a/wren-core-legacy/wren-base/pom.xml
+++ b/wren-core-legacy/wren-base/pom.xml
@@ -18,7 +18,7 @@
     <parent>
         <groupId>io.wren</groupId>
         <artifactId>wren-root</artifactId>
-        <version>0.17.1</version>
+        <version>0.17.2-SNAPSHOT</version>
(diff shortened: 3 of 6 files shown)


douenergy:
Thanks @loaferxuan for the fix! The change looks readable to me. However, we’ve switched the underlying type system to PyArrow, so the CI failed because of that. Could you rebase onto the main branch?
Welcome to the Wren Engine community!

PR #1235, opened 2025-06-27
fix(ibis): using DataFusion to format the query result

goldmedal:
# Description
Because DuckDB doesn't support setting the timezone offset as the session timezone, we need to change to DataFusion to format the query result. There are some changes to the result:
- The timezone will be shown as the offset for the result always.
  - `2011-01-01 12:00:00 UTC` -> `2011-01-01 12:00:00 +00:00`
- `TIMESTAMP` and `DATE` types won't be affected by the session timezone. Session timezone only affects the  `TIMESTAMP WITH TIMEZONE` type.

# Known Issue
- MySQL can't recognize the timestamp and timestamp with timezone. We show them with timezone always. 
- Snowflake supports the timestamp tz literal using the offset only.

## Summary by CodeRabbit

* **New Features**
  * Switched SQL execution backend from DuckDB to DataFusion, improving performance and compatibility.
  * Added session timezone support for query results, ensuring timestamp fields reflect the correct timezone.

* **Bug Fixes**
  * Updated timestamp and interval formatting in query results for consistency across data sources.
  * Improved handling and formatting of numeric and floating-point values in query outputs.

* **Chores**
  * Updated type annotations to modern Python syntax.
  * Added DataFusion as a new dependency.
  * Enhanced test cases to reflect changes in data formatting and timezone representation.
  * Modernized string formatting style in core components for improved code clarity.


goldmedal:
BigQuery test has been tested locally. Let's ignore CI failure.

douenergy (approved):

PR #1243, opened 2025-07-02
feat(ibis): Add safe rounding for Decimal columns to support PyArrow Decimal128 limits

douenergy:
We ensures decimal columns are cast and rounded to a fixed scale (default: 9) with a maximum precision of 36, aligning with the pyarrow Decimal128 limit (max 38). 

It prevents overflow or unsupported precision issues when exporting or operating on decimal columns using pyarrow backends.

### DIFF
--- a/ibis-server/app/model/connector.py
+++ b/ibis-server/app/model/connector.py
@@ -8,7 +8,6 @@ from typing import Any
 import ibis
 import ibis.expr.datatypes as dt
 import ibis.expr.schema as sch
-import ibis.formats
 import pandas as pd
 import pyarrow as pa
 import sqlglot.expressions as sge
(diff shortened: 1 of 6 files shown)


ibis-server/app/model/connector.py
```diff
@@ -190,7 +194,7 @@ def query(self, sql: str, limit: int) -> pa.Table:
 
 class DuckDBConnector:
     def __init__(self, connection_info: ConnectionInfo):
-        import duckdb  # noqa: PLC0415
```
goldmedal: Why remove this comment?

ibis-server/app/query_cache/__init__.py
```diff
@@ -27,9 +29,11 @@ def get(self, data_source: str, sql: str, info) -> Any | None:
         if op.exists(cache_file_name):
             try:
                 logger.info(f"Reading query cache {cache_file_name}")
-                df = ibis.read_parquet(full_path).to_pyarrow()
+                ibis_table = ibis.read_parquet(full_path)
+                ibis_table = round_decimal_columns(ibis_table)
```
goldmedal: We have rounded the query result. Is it required for the cache result?

goldmedal (approved):
Thanks @douenergy 👍

PR #1245, opened 2025-07-04
fix(core): fix RLAC for model with an alias

### LINKED ISSUE #1222: Alias is not converted when row level access control is configured
**Describe the bug**

When I setup the row level access control at the same time setup the table with alias, the translated planned SQL does not convert the table name inside the row level access control condition. 

**To Reproduce**

1. Setup RACL with for example nationkey = @session_user
2. Define the query like below:

```sql

SELECT * FROM customer ordered LIMIT 20;

```

You will get error `customer.nationkey` not found. It indicates that the key used in RACL was not converted correctly with alias.

I tested the latest version of ibis-server.

goldmedal:
- close #1222 

DataFusion plan a filter in the subquery to the plan like
```
SELECT int_col, double_col FROM (SELECT int_col, double_col, CAST(date_string_col as VARCHAR) FROM alltypes_plain where int_col > 1) d
-----
Logical Plan: Projection: d.int_col, d.double_col
  SubqueryAlias: d
    Projection: alltypes_plain.int_col, alltypes_plain.double_col, CAST(alltypes_plain.date_string_col AS Utf8View)
      Filter: alltypes_plain.int_col > Int64(1)
        TableScan: alltypes_plain
```

We need to follow the pattern to build our custom plan for datafusion unparsing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of row-level security (RLS) filters to ensure correct query generation and aliasing, aligning with expected SQL behavior.

* **Tests**
  * Updated test snapshots to reflect changes in the structure of queries with RLS filters.
  * Added a new test to verify correct SQL generation for models with RLS filters and table aliases.

### DIFF
--- a/wren-core-py/tests/test_modeling_core.py
+++ b/wren-core-py/tests/test_modeling_core.py
@@ -315,7 +315,7 @@ def test_rlac():
     rewritten_sql = session_context.transform_sql(sql)
     assert (
         rewritten_sql
-        == "SELECT customer.c_custkey, customer.c_name FROM (SELECT customer.c_custkey, customer.c_name FROM (SELECT __source.c_custkey AS c_custkey, __source.c_name AS c_name FROM main.customer AS __source) AS customer) AS customer WHERE customer.c_name = 'test_user'"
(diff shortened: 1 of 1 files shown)


douenergy:
Thanks @goldmedal

PR #1274, opened 2025-07-29
fix(core): fix count statement for BigQuery

goldmedal:
# Description

We fiedx the alias name of BigQuery in https://github.com/Canner/datafusion/pull/1
However, we didn't apply it in our inner dialect.


douenergy (approved):

douenergy:
Thanks @goldmedal

PR #1271, opened 2025-07-24
feat(ibis): Introduce bigquery white function list

douenergy:
### Summary
Implements automated validation for BigQuery functions to prevent runtime errors from unsupported functions.
### Changes

`ibis-server/resources/fuzzing/bigquery.json` - 311 SQL test cases for comprehensive function coverage
`ibis-server/tools/white_remote_function.py` - Validation tool that categorizes functions as:

✅ Supported
❌ Unsupported (function not found)
🔧 Syntax errors (need SQL fixes)

## Summary by CodeRabbit

* **New Features**
  * Added a comprehensive set of 311 BigQuery SQL function test cases for improved coverage.
  * Introduced a new script to automate testing of SQL queries against a remote API, with detailed error categorization and reporting.
  * Enabled function retrieval from a whitelist CSV for specified data sources.
  * Added support for configuring and using a remote whitelist path via environment variables.

* **Bug Fixes**
  * Updated test expectations and descriptions for BigQuery function listing to reflect current results.

* **Chores**
  * Adjusted configuration and test resource paths to support new function list management.

PR #1275, opened 2025-07-30
feat(ibis): remove `to_char` from BigQuery white function list

douenergy:
following https://github.com/Canner/wren-engine/pull/1271
related to https://github.com/tobymao/sqlglot/issues/5496 

When performing fuzz testing, the test case for to_char is
```sql 
SELECT to_char(current_date(), 'YYYY-MM-DD') AS result
```
and sqlglot is smart enough to rewrite it to:

```sql 
SELECT CAST(CURRENT_DATE AS STRING) AS `result`
```

However, when combining to_char with the cast function, the rewrite does not work:
```sql 
SELECT TO_CHAR(
    CAST(contacts.createdAt AS TIMESTAMP WITH TIME ZONE), 
    'YYYY-MM'
) AS year_month,
COUNT(*) AS contact_count
FROM contacts
WHERE CAST(contacts.createdAt AS TIMESTAMP WITH TIME ZONE) >= CAST('2023-01-01 00:00:00' AS TIMESTAMP WITH TIME ZONE)
  AND CAST(contacts.createdAt AS TIMESTAMP WITH TIME ZONE) < CAST('2025-01-01 00:00:00' AS TIMESTAMP WITH TIME ZONE)
GROUP BY year_month
ORDER BY year_month ASC
```

This results in the error:

Argument 'format' is not supported for expression 'ToChar' when targeting Dialect.

## Summary by CodeRabbit

* **Tests**
  * Updated a test to expect a different number of functions returned from the `/functions` endpoint.


douenergy:
I have run all bigquery test on my local
bigquery

PR #1276, opened 2025-08-05
feat(snowflake): add optional warehouse support to connection

douenergy:
We should not restrict users from specifying the warehouse they use. If the warehouse is None, Snowflake will use the default warehouse automatically.

### DIFF
--- a/ibis-server/app/model/data_source.py
+++ b/ibis-server/app/model/data_source.py
@@ -348,25 +348,37 @@ class DataSourceExtension(Enum):
 
     @staticmethod
     def get_snowflake_connection(info: SnowflakeConnectionInfo) -> BaseBackend:
+        # private key authentication
         if hasattr(info, "private_key") and info.private_key:
-            return ibis.snowflake.connect(
-                user=info.user.get_secret_value(),
-                account=info.account.get_secret_value(),
-                database=info.database.get_secret_value(),
-                schema=info.sf_schema.get_secret_value(),
-                warehouse=info.warehouse.get_secret_value(),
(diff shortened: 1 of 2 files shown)


ibis-server/app/model/data_source.py
```diff
@@ -348,25 +348,37 @@ def get_oracle_connection(info: OracleConnectionInfo) -> BaseBackend:
...
+                "database": info.database.get_secret_value(),
+                "schema": info.sf_schema.get_secret_value(),
+            }
+
+            # warehouse if it exists and is not None/empty
+            if hasattr(info, "warehouse") and info.warehouse:
+                connection_params["warehouse"] = info.warehouse.get_secret_value()
+            if info.kwargs:
+                connection_params.update(info.kwargs)
+            return ibis.snowflake.connect(**connection_params)
```
goldmedal: It could be simplified to the following code, then we don't need to set warehouse at L360~L365
```suggestion
            # password authentication
            connection_params = {
                "user": info.user.get_secret_value(),
                "password": info.password.get_secret_value(),
                "account": info.account.get_secret_value(),
                "database": info.database.get_secret_value(),
                "schema": info.sf_schema.get_secret_value(),
            }

        # warehouse if it exists and is not None/empty
        if hasattr(info, "warehouse") and info.warehouse:
            connection_params["warehouse"] = info.warehouse.get_secret_value()
        if info.kwargs:
            connection_params.update(info.kwargs)
        return ibis.snowflake.connect(**connection_params)
```

goldmedal (approved):
Thanks @douenergy. I have only one suggestion for the get_snowflake_connection

PR #1277, opened 2025-08-07
feat(ibis): Include HTTP headers in query cache key generation

douenergy:
Since we introduce access control and query timeout ... , query cache now includes headers and timeout settings in cache keys to prevent incorrect data sharing.

## Summary by CodeRabbit

* **New Features**
  * Query caching now considers specific HTTP headers, including Wren variable headers and statement timeout, to enhance cache precision based on request context.

* **Bug Fixes**
  * Irrelevant headers no longer impact cache keys, reducing unnecessary cache misses.
  * Improved error handling when fallback is disabled, returning appropriate error responses.

* **Tests**
  * Added extensive tests validating cache behavior with various relevant and irrelevant headers, confirming accurate cache hits and misses.
  * Updated tests to reflect fallback behavior changes and error responses when fallback is disabled.

### DIFF
--- a/ibis-server/app/query_cache/__init__.py
+++ b/ibis-server/app/query_cache/__init__.py
@@ -17,8 +17,10 @@ class QueryCacheManager:
         self.root = root
 
     @tracer.start_as_current_span("get_cache", kind=trace.SpanKind.INTERNAL)
-    def get(self, data_source: str, sql: str, info) -> Any | None:
-        cache_key = self._generate_cache_key(data_source, sql, info)
+    def get(
+        self, data_source: str, sql: str, info, headers: dict[str, str] | None = None
+    ) -> Any | None:
+        cache_key = self._generate_cache_key(data_source, sql, info, headers)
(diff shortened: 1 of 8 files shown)


ibis-server/app/query_cache/__init__.py
```diff
@@ -75,14 +80,50 @@ def get_cache_file_timestamp(self, data_source: str, sql: str, info) -> int | No
...
+    ) -> str:
+        if not headers:
+            return ""
+
+        # Define which headers should be included in cache key
+        # These are headers that can affect the query results
+        cache_relevant_headers = [
+            "x-wren-variable",  # All Wren variable headers (prefix match)
+            "x-wren-db-statement_timeout",
+            "x-wren-timezone",
```
goldmedal: It's better to import their variables from `ibis-server/app/dependencies.py` instead of using the value directly.

goldmedal (approved):
Thanks @douenergy. 👍

PR #1279, opened 2025-08-12
feat(core): add BigQuery EXTRACT function support and window frame validation

goldmedal:
# Description
- DataFusion planner will plan a `EXTRACT` statement to a function call `date_part`. However, it's not a valid function for BigQuery. We should unparse the `date_part` function back to `EXTRACT` statement
- BigQuery only allows [the aggregation to have the window frame clause](https://cloud.google.com/bigquery/docs/reference/standard-sql/window-function-calls#def_window_frame). Use `window_func_support_window_frame` to check if a window function call should generate the window frame clause. 

## Summary by CodeRabbit

- New Features
  - BigQuery: date_part now emits EXTRACT(field FROM expr) SQL with broader field-name support and stricter argument validation.
  - BigQuery: adjusted window-function framing behavior for several ranking/analytic functions to match BigQuery semantics.

- Refactor
  - Safer, more idiomatic internal logic and a minor signature/lifetime clarification (no behavioral change).

- Tests
  - Added BigQuery EXTRACT and window-function round-trip tests and minor test cleanups.

### DIFF
--- a/wren-core-base/src/mdl/utils.rs
+++ b/wren-core-base/src/mdl/utils.rs
@@ -43,7 +43,7 @@ pub(crate) fn parse_identifiers_normalized(
     })
 }
 
-pub fn quote_identifier(s: &str) -> Cow<str> {
+pub fn quote_identifier(s: &str) -> Cow<'_, str> {
     if needs_quotes(s) {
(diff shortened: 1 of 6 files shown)


douenergy:
BigQuery will show an error message like: `Window framing clause is not allowed for analytic function ROW_NUMBER`.

PR #1283, opened 2025-08-14
fix(core): add quotes for the model subquery alias

goldmedal:
I missed to add quotes for the alias of the model subquery. It causes some case-sensitive issues.

## Summary by CodeRabbit

- New Features
  - Improved support for Unicode model and column names in generated SQL.

- Bug Fixes
  - Model aliases are now properly quoted to handle special or non-ASCII characters.
  - More reliable combination and application of row-level access filters in query generation.

- Tests
  - Added and adjusted tests covering Unicode identifiers, row-level access control, and related behavior.


douenergy (approved):

douenergy:
Thanks @goldmedal

PR #1285, opened 2025-08-14
fix(ibis): Increase Redshift get_table_list limit

JakeCowton:
Setting a limit of 500 means that, if you have more than 500 total columns across all tables, any further columns will silently be ignored. This issue is made worse by the fact that subsequent queries to check for schema updates won't necessarily return the columns in the same order, which means different columns might be ignored which are detected as columns being deleted.

This change increases the limit to 50,000. It doesn't _solve_ the issue, but makes it highly unlikely to happen.

### DIFF
--- a/ibis-server/app/model/metadata/redshift.py
+++ b/ibis-server/app/model/metadata/redshift.py
@@ -51,7 +51,7 @@ class RedshiftMetadata(Metadata):
         # we reuse the connector.query method to execute the SQL
         # so we have to give a limit to avoid too many rows
         # the default limit for tables metadata is 500, i think it's a sensible limit
-        response = self.connector.query(sql, limit=500).to_pylist()
+        response = self.connector.query(sql, limit=50000).to_pylist()
 
         unique_tables = {}
         for row in response:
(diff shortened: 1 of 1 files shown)


ibis-server/app/model/metadata/redshift.py
```diff
@@ -104,7 +104,7 @@ def get_constraints(self) -> list[Constraint]:
         # we reuse the connector.query method to execute the SQL
         # so we have to give a limit to avoid too many rows
         # the default limit for constraints metadata is 500, i think it's a sensible limit
-        response = self.connector.query(sql, limit=500).to_pylist()
+        response = self.connector.query(sql, limit=50000).to_pylist()
```
goldmedal:
```suggestion
        response = self.connector.query(sql).to_pylist()
```
An alternative is setting no limit for this query. This SQL has only 9 columns. I think it's affordable to accept more than 50k rows.

goldmedal (approved):
Thanks @JakeCowton for the contribution. It makes sense to me. I only have one comment for the limit setting but isn't required.

goldmedal (approved):
Thanks @JakeCowton. LGTM

ibis-server/app/model/metadata/redshift.py
JakeCowton: Thanks @goldmedal , I'd assumed based on the comments above the query call that `limit` was a required field. I've removed the limit from both queries and cleaned up the comments suggesting a limit is required.

PR #1331, opened 2025-09-24
fix(core): normalize the name of session property

goldmedal:
# Description
The name of a session property should be case-insensitive. This PR adds the field `normalized_name` for `SessionProperty` which is used to match the required property.


douenergy (approved):

douenergy:
Thanks @goldmedal

PR #1340, opened 2025-10-03
Fix: MSSQL add N char

ansutung:


goldmedal:
@ansutung, could you rebase on the latest main and add the description for what you want to propose?

PR #1345, opened 2025-10-08
feat(ibis): Show full Redshift table metadata by querying system catalogs

douenergy:
The current Redshift metadata query uses information_schema.tables and information_schema.columns, which only shows schemas that are in the user's search_path or owned by the user. This causes visibility issues like https://github.com/Canner/WrenAI/issues/1953


goldmedal:
@douenergy, is this PR tested well?

goldmedal:
@douenergy  did you check if a user with a lower permission could get the tables that he shouldn't see through the pg_catalog path?

PR #1359, opened 2025-10-29
fix(ibis): add support for ClickHouse wrapper types (Nullable, Array, LowCardinality)

rishabh1815769:
**Summary**
Fixes type casting issues for ClickHouse wrapper types such as Nullable(), Array(), and LowCardinality() by extracting the inner type before mapping to Rust Wren Engine column types.

**Problem**
Previously, ClickHouse columns with wrapper types like Nullable(String), Array(Int32) etc were not being properly recognized and would be marked as UNKNOWN type, causing issues with type inference and query execution.

## Summary by CodeRabbit

* **New Features**
  * Added support for datetime64 column type in ClickHouse metadata.

* **Bug Fixes**
  * Improved handling of wrapped ClickHouse data types (Array, Nullable, LowCardinality).
  * Fixed column nullability detection based on data type wrappers.

### DIFF
--- a/ibis-server/app/model/metadata/clickhouse.py
+++ b/ibis-server/app/model/metadata/clickhouse.py
@@ -31,6 +31,7 @@ CLICKHOUSE_TYPE_MAPPING = {
     # Date/Time Types
     "date": RustWrenEngineColumnType.DATE,
     "datetime": RustWrenEngineColumnType.TIMESTAMP,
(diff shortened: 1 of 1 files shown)


goldmedal:
Thanks @rishabh1815769 for working on this. There is a ruff check failure. You can fix it using our formatting command.
- Install [just](https://github.com/casey/just)
- Execute `just fmt` at `ibis-server`

PR #1563, opened 2026-04-16
fix(ci): update Cargo.lock in wren-core-py release workflow

goldmedal:
## Summary
- Maturin v1.13+ uses `--locked` by default when a `Cargo.lock` exists. The "Set version" step was updating `Cargo.toml` and `pyproject.toml` but not `Cargo.lock`, causing all wheel builds to fail with a lockfile mismatch error.
- Extends the version-setting Python script to also patch the `wren-core-py` version in `Cargo.lock`, in both the `build-wheels` and `build-sdist` jobs.

Fixes https://github.com/Canner/wren-engine/actions/runs/24493639398

## Test plan
- [ ] Re-run the wren-core-py release after merging to verify all 4 wheel builds (Linux, macOS Intel, macOS ARM, Windows) succeed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Updated the release workflow to synchronize version information across all configuration files during package publishing.

### DIFF
--- a/.github/workflows/publish-wren-core-py.yml
+++ b/.github/workflows/publish-wren-core-py.yml
@@ -55,6 +55,13 @@ jobs:
               text = open(path).read()
               text = re.sub(r'^(version\s*=\s*)".*?"', rf'\1"{version}"', text, count=1, flags=re.MULTILINE)
               open(path, "w").write(text)
+          lock = open("wren-core-py/Cargo.lock").read()
+          lock = re.sub(
+              r'(name = "wren-core-py"\nversion = )".*?"',
+              rf'\1"{version}"',
+              lock, count=1,
+          )
+          open("wren-core-py/Cargo.lock", "w").write(lock)
       - name: Build wheel
         uses: PyO3/maturin-action@v1
         with:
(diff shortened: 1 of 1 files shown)


douenergy (approved):

.github/workflows/publish-wren-core-py.yml
```diff
@@ -55,6 +55,13 @@ jobs:
               text = open(path).read()
               text = re.sub(r'^(version\s*=\s*)".*?"', rf'\1"{version}"', text, count=1, flags=re.MULTILINE)
               open(path, "w").write(text)
+          lock = open("wren-core-py/Cargo.lock").read()
+          lock = re.sub(
+              r'(name = "wren-core-py"\nversion = )".*?"',
+              rf'\1"{version}"',
```
utafrali, line 61: The regex `r'(name = "wren-core-py"\nversion = )".*?"'` works correctly here because `\n` inside a raw string is still interpreted as a newline by the regex engine. Worth noting for anyone who reads this later: unlike the existing `Cargo.toml`/`pyproject.toml` substitution above which uses `re.MULTILINE` with a `^` anchor, this pattern crosses a line boundary explicitly via `\n`, so `re.MULTILINE` is intentionally absent. A short inline comment (`# \n matches literal newline between name and version fields`) would make the intent self-evident.

utafrali (approved):
This is a minimal, targeted fix for a real breakage introduced by maturin v1.13+ defaulting to `--locked`. The regex correctly identifies the wren-core-py entry in Cargo.lock, `count=1` prevents spurious replacements, and the fix is applied consistently in both the wheel and sdist jobs. Ready to merge.

